PR #11, opened 2018-06-07
Changes for release v0_1

jradcliff:


jradcliff:
I'll update the change log and pom versions in a subsequent PR.

PR #12, opened 2018-06-07
Modify pom versions and change log for release 0.2.0.

jradcliff:


jradcliff:
These changes put poms in a state expected by the Maven release plugin.

PR #13, opened 2018-06-18
Changes for release v0_2

jradcliff:
- Regenerate from the latest protos.
- Updates to Hotel Ads examples for PercentCpc bidding strategy.


google-ads-examples/pom.xml
```diff
@@ -55,7 +55,7 @@
     <dependency>
       <groupId>com.google.api-ads</groupId>
       <artifactId>google-ads</artifactId>
-      <version>0.2.0</version>
+      <version>0.2.1-SNAPSHOT</version>
```
mcloonan: I think you want an actual version number, not a SNAPSHOT.

PR #34, opened 2018-09-27
Add ResourceNames utility.

jradcliff:
Includes a few samples of updated examples to demonstrate usage.

### DIFF
--- a/google-ads/src/main/java/com/google/ads/googleads/lib/utils/ResourceNames.java
+++ b/google-ads/src/main/java/com/google/ads/googleads/lib/utils/ResourceNames.java
@@ -1,0 +1,191 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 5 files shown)


google-ads/src/main/java/com/google/ads/googleads/lib/utils/ResourceNames.java
```diff
@@ -0,0 +1,191 @@
...
+   * standard delimiter for composite identifiers.
+   */
+  private static String concatLongs(long... components) {
+    return IDENTIFIER_JOINER.join(Arrays.stream(components).mapToObj(Long::toString).iterator());
+  }
+
+  /** Returns the account budget proposal resource name for the specified components. */
+  public static String accountBudgetProposal(long customerId, long accountBudgetProposalId) {
+    return AccountBudgetProposalName.format(
+        Long.toString(customerId), concatLongs(accountBudgetProposalId));
```
mcloonan: Why do you occasionally use "concatLongs" on individual IDs? Should these be Long.toString as well?

jradcliff: Good catch. I've updated all of the single component cases to just use Long.toString.

PR #40, opened 2018-11-03
Add note to pom.xml to trigger presubmit

jradcliff:
Attempt 2 with Write access only for `kokoro-team`.


jradcliff:
Test for presubmit

PR #71, opened 2019-03-14
Add deserialize of GoogleAdsFailure message to logging

nwbirnie:
Change-Id: I23f7d004e277acb20284e338c4d1d2fb8bd72c6d

### DIFF
--- a/google-ads/src/main/java/com/google/ads/googleads/lib/BaseGoogleAdsException.java
+++ b/google-ads/src/main/java/com/google/ads/googleads/lib/BaseGoogleAdsException.java
@@ -116,7 +116,10 @@ public abstract class BaseGoogleAdsException extends ApiException {
 
     /** Create an empty GoogleAdsFailure instance for this version. */
     @VisibleForTesting
-    public abstract U createDefaultFailure();
+    public abstract U createGoogleAdsFailure();
+
+    public abstract U createGoogleAdsFailure(byte [] serializedBytes)
+        throws InvalidProtocolBufferException;
 
     protected static Metadata.Key<byte[]> createKey(String trailerKey) {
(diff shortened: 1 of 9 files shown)


google-ads/src/main/java/com/google/ads/googleads/lib/logging/Event.java
```diff
@@ -128,6 +134,29 @@ public String getResponseAsText() {
       return getResponse() == null ? null : getResponse().toString();
     }
 
+    /**
+     * Attempts to read the GoogleAdsFailure from the response, if present. This is a relatively
+     * expensive operation when the proto is present, so take care not to negatively impact
```
jradcliff: Instead of:

> so take care not to negatively impact performance

which  could be interpreted as a note for the maintainers of this method, how about:

> so call this method judiciously.

google-ads/src/main/java/com/google/ads/googleads/lib/BaseGoogleAdsException.java
```diff
@@ -116,7 +116,10 @@ protected abstract T createException(
 
     /** Create an empty GoogleAdsFailure instance for this version. */
     @VisibleForTesting
-    public abstract U createDefaultFailure();
+    public abstract U createGoogleAdsFailure();
```
jradcliff: Please add a 2nd reviewer

google-ads/src/main/java/com/google/ads/googleads/lib/logging/Event.java
nwbirnie: Done

jradcliff:
@AnashOommen , could you PTAL?

Thanks

PR #91, opened 2019-04-22
Examples/add dynamic page feed

devchas:


google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
```diff
@@ -0,0 +1,543 @@
...
+    // Issue the search request.
+    try (GoogleAdsServiceClient googleAdsServiceClient =
+        googleAdsClient.getLatestVersion().createGoogleAdsServiceClient()) {
+      SearchPagedResponse searchPagedResponse = googleAdsServiceClient.search(request);
+      // Get the first result because we only need the single feed item we created previously.
+      GoogleAdsRow googleAdsRow = searchPagedResponse.getPage().getResponse().getResults(0);
+      // Get the attributes list from the feed and create a map with keys of each attribute and
+      // values of each corresponding ID.
+      List<FeedAttribute> feedAttributeList = googleAdsRow.getFeed().getAttributesList();
+      Map<String, Long> feedAttributeMap = new Hashtable<>();
```
jradcliff: Please use HashMap instead. Hashtable is left over from the very early days of Java and is rarely used these days.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
```diff
@@ -0,0 +1,543 @@
...
+              googleAdsRow.getFeed().getResourceName(),
+              feedAttributeMap.get("Page URL"),
+              feedAttributeMap.get("Label"));
+
+      return feedDetails;
+    }
+  }
+
+  /**
+   * Creates a field mapping for a given feed.
```
jradcliff: "field" => "feed"

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
```diff
@@ -0,0 +1,543 @@
...
+            .setFeedAttributeId(Int64Value.of(feedDetails.getUrlAttributeId()))
+            .setDsaPageFeedField(DsaPageFeedCriterionField.PAGE_URL)
+            .build();
+    AttributeFieldMapping labelFieldMapping =
+        AttributeFieldMapping.newBuilder()
+            .setFeedAttributeId(Int64Value.of(feedDetails.getLabelAttributeId()))
+            .setDsaPageFeedField(DsaPageFeedCriterionField.LABEL)
+            .build();
+
+    // Map the FeedAttributeIds to the fieldId constants.
```
jradcliff: Please update the comment. Looks like this was inadvertently copied from line 287.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
```diff
@@ -0,0 +1,543 @@
...
+      long customerId,
+      FeedDetails feedDetails,
+      String dsaPageUrlLabel) {
+    List<String> urls =
+        Arrays.asList(
+            "http://www.example.com/discounts/rental-cars",
+            "http://www.example.com/discounts/hotel-deals",
+            "http://www.example.com/discounts/flight-deals");
+
+    List<FeedItemOperation> ops = new ArrayList<>();
```
jradcliff: nit: please use full name 'operations'

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
```diff
@@ -0,0 +1,543 @@
...
+            .setPageSize(PAGE_SIZE)
+            .setQuery(query)
+            .build();
+
+    // Create the service client.
+    try (GoogleAdsServiceClient googleAdsServiceClient =
+        googleAdsClient.getLatestVersion().createGoogleAdsServiceClient()) {
+      SearchPagedResponse searchPagedResponse = googleAdsServiceClient.search(request);
+
+      // Throw an exception a campaign with the provided ID does not exist.
```
jradcliff: "a campaign" => "if a campaign"

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
```diff
@@ -0,0 +1,543 @@
...
+              .getResponse()
+              .getResults(0)
+              .getCampaign()
+              .getDynamicSearchAdsSetting();
+
+      // Get the campaign domain name.
+      String domainName = dynamicSearchAdsSetting.getDomainName().getValue();
+
+      // Throw an exception if the campaign is not a DSA campaign.
+      if (domainName == null || domainName == "") {
```
jradcliff: Instead of checking the value in this way, use `dynamicSearchAdsSetting.hasDomainName()`. The library will never return `null` from `StringValue.getValue()`, but `hasDomainName()` will properly check for null.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
```diff
@@ -0,0 +1,543 @@
...
+      for (MutateCampaignResult mutateCampaignResult : response.getResultsList()) {
+        System.out.printf(
+            "Updated campaign with resourceName: %s.%n", mutateCampaignResult.getResourceName());
+      }
+    }
+    return;
+  }
+
+  /**
+   * Returns the DSA settings for a campaign. Throws an error if the campaign does not exists or is
```
jradcliff: "exists" => "exist"

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
```diff
@@ -0,0 +1,543 @@
...
+      if (domainName == null || domainName == "") {
+        throw new IllegalArgumentException(
+            "Campaign with ID " + campaignId + " is not a DSA campaign.");
+      }
+      return dynamicSearchAdsSetting;
+    }
+  }
+
+  /**
+   * Updates an ad group criterion targeting the DSA label.
```
jradcliff: Instead of "Updates", indicate this "Creates" an ad group criterion.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
```diff
@@ -0,0 +1,543 @@
...
+   * @param customerId the client customer ID in which to create criterion.
+   * @param adGroupId the adGroup ID.
+   * @param campaignId the campaign ID.
+   * @throws GoogleAdsException if an API request failed with one or more service errors.
+   */
+  private void runExample(
+      GoogleAdsClient googleAdsClient, long customerId, long campaignId, long adGroupId) {
+    String dsaPageUrlLabel = "discounts";
+
+    // Get the page feed details. This code example creates a new feed, but you can
```
jradcliff: I realize we're not always consistent on this, but the Javadoc guide indicates that comments should be written with the assumed prefix of "This line...", "This method...", etc.

So this would be "Gets" instead of "Get".

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
devchas: Done.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
devchas: Done.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
devchas: Done.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
devchas: Done.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
devchas: Done.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
devchas: Done.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
devchas: Done.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
devchas: Done.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
devchas: Done.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
```diff
@@ -0,0 +1,540 @@
...
+              googleAdsRow.getFeed().getResourceName(),
+              feedAttributeMap.get("Page URL"),
+              feedAttributeMap.get("Label"));
+
+      return feedDetails;
+    }
+  }
+
+  /**
+   * Creates a feed mapping for a given feed.
```
nwbirnie: Can we refer to the guide for more information here? Feed Mapping is kind of esoteric - it's needed because Feeds can have arbitrary structure and we need to know how these map to the placeholder types, but a new user reading this wouldn't have context.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
```diff
@@ -0,0 +1,540 @@
...
+   *
+   * @param googleAdsClient the Google Ads API client.
+   * @param customerId the client customer ID in which to create criterion.
+   * @param campaignId the campaign ID.
+   */
+  private DynamicSearchAdsSetting getDsaSetting(
+      GoogleAdsClient googleAdsClient, long customerId, long campaignId) {
+    // Creates the query.
+    // You must request all DSA fields in order to update the DSA settings in the following step.
+    String query =
```
nwbirnie: Why do we need this round trip? Can't we just overwrite the existing DSA setting?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
```diff
@@ -0,0 +1,540 @@
...
+      MutateAdGroupCriteriaResponse mutateAdGroupCriteriaResponse =
+          adGroupCriterionServiceClient.mutateAdGroupCriteria(
+              Long.toString(customerId), ImmutableList.of(operation));
+      // Displays the results.
+      for (MutateAdGroupCriterionResult result : mutateAdGroupCriteriaResponse.getResultsList()) {
+        System.out.printf(
+            "Created ad group criterion with resource name '%s'.%n", result.getResourceName());
+      }
+    }
+    return;
```
nwbirnie: Nit: remove this return statement, not required.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
```diff
@@ -0,0 +1,540 @@
...
+            .setName(StringValue.of("Label"))
+            .build();
+
+    // Creates the feed.
+    Feed feed =
+        Feed.newBuilder()
+            .setName(StringValue.of("DSA Feed #" + System.currentTimeMillis()))
+            .addAttributes(urlAttribute)
+            .addAttributes(labelAttribute)
+            .setOrigin(FeedOrigin.USER)
```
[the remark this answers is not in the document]
devchas: Confirmed this is not required. I have removed.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
```diff
@@ -0,0 +1,540 @@
...
+        googleAdsClient.getLatestVersion().createGoogleAdsServiceClient()) {
+      SearchPagedResponse searchPagedResponse = googleAdsServiceClient.search(request);
+      // Gets the first result because we only need the single feed item we created previously.
+      GoogleAdsRow googleAdsRow = searchPagedResponse.getPage().getResponse().getResults(0);
+      // Gets the attributes list from the feed and create a map with keys of each attribute and
+      // values of each corresponding ID.
+      List<FeedAttribute> feedAttributeList = googleAdsRow.getFeed().getAttributesList();
+      HashMap<String, Long> feedAttributeMap = new HashMap<>();
+      for (FeedAttribute feedAttribute : feedAttributeList) {
+        feedAttributeMap.put(feedAttribute.getName().getValue(), feedAttribute.getId().getValue());
```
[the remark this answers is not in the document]
devchas: No, you cannot set the ID.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
```diff
@@ -0,0 +1,540 @@
...
+    @Parameter(names = ArgumentNames.AD_GROUP_ID, required = true)
+    private long adGroupId;
+  }
+
+  private static class FeedDetails {
+    private String feedResourceName;
+    private long urlAttributeId;
+    private long labelAttributeId;
+
+    private FeedDetails(String feedResourceName, long urlAttributeId, long labelAttributeId) {
```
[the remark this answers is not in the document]
devchas: I think it's a useful convenience because the feed attributes are returned as a list containing a name and ID. That means you would have to loop through that list and search for the "Page URL" and "Label" names in order to get the associated IDs in both the createFeedMapping and createFeedItems methods. Within each of those methods, you'd either have to loop through the attributes twice (once for each attribute) or create a map in each method to identify which ID belongs to which name. I think the current implementation is cleaner and easier.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
devchas: This guide does not yet exist for GAAPI, but I think we should it one once it is created.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
devchas: Yes, you can overwrite the existing DSA settings. The potential downside is that you have to know the values for domain_name, language_code, and use_supplied_urls_only you used when originally creating the campaign if you want them to remain the same. Presumably you would want the values to remain the same because you set those values initially. It seems counter-intuitive to me to have to look those up or introduce the risk of potentially making an inadvertent change to the DB. In my opinion that inconvenience/risk outweighs the cost of the additional roundtrip, so I'll leave it as for now. That said, it's an easy fix, so LMK if you want to disagree, and I can change it.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddDynamicPageFeed.java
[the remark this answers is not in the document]
jradcliff: How about this instead?

1) Have `createFeed` return the Feed resource name.
2) Have `getFeed` take the Feed resource name as a parameter and return a `Map<DsaPageFeedCriterionField, FeedAttribute>`.
2) Have `createFeedMapping` take the Feed resource name and a `Map<DsaPageFeedCriterionField, FeedAttribute>` as parameters, and return a `Map<DsaPageFeedCriterionField, AttributeFieldMapping>'.
3) Have `createFeedItem` take the Feed resource name and a `Map<DsaPageFeedCriterionField, AttributeFieldMapping>' as parameters.

The benefit of this approach is that it uses API types and thereby helps users gain an understanding of what those types represent and contain. It also uses the `DsaPageCriterionField` enum values in place of names.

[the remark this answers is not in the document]
devchas: SGTM, updated.

devchas:
Please hold off on merging this PR until further notice. There is a bug associated with adding FeedItemAttributes for DSA feeds.

nwbirnie:
Perhaps this PR been superceded? Could we close it?

devchas:
Yes, this is old. Closing.

PR #98, opened 2019-05-01
Changes for release v1_2.

devchas:


nwbirnie:
Poking CI again with webhooks enabled for comments

PR #115, opened 2019-05-22
Added getter for version name

Raibaz:


nwbirnie:
LGTM, do you want to also include updating the visibility modifier on ApiCatalog.getDefault()?

Raibaz:
Don't think it's needed, the ApiCatalog interface is public so its methods are public by default.

google-ads/src/main/java/com/google/ads/googleads/lib/catalog/Version.java
```diff
@@ -50,6 +50,14 @@
         versionName.matches("v[0-9]+"), "Invalid version name: %", versionName);
   }
 
+  /**
+   * Returns the version name, e.g. "v1".
+   * @return the version name
+   */
+  public String getVersionName() {
```
jradcliff: Could you elaborate on the motivation behind this change?

Raibaz: Accessing the name (i.e. the number) of the latest version programmatically may be useful for client applications that want to use reflection to instantiate classes like resources or enums, like an entity class generator that generates a class for each resource and wants to use the appropriate class for the enum fields.

nwbirnie:
LGTM

PR #119, opened 2019-05-23
Examples/handle rate exceeded error

devchas:


google-ads-examples/src/main/java/com/google/ads/googleads/examples/errorhandling/HandleRateExceededError.java
```diff
@@ -0,0 +1,244 @@
...
+   * @param adGroupId the ID of the ad group to which keywords are added.
+   */
+  private void runExample(GoogleAdsClient googleAdsClient, long customerId, long adGroupId)
+      throws InterruptedException {
+    final int NUM_THREADS = 500;
+
+    List<Thread> threads = new ArrayList<>();
+
+    // Adds 500 threads that validate keywords to a list.
+    for (int i = 0; i < NUM_THREADS; i++) {
```
jradcliff: I'm concerned about an example that intentionally tries to exceed rate limits. Could you change this to just demonstrate the error handling/retry approach, but without the intentional triggering of limits?

Issue #112, opened 2019-05-20
Request/Response logging for partial failures  is in 'DEBUG' level instead of 'INFO'


YaelWeizer:
Request/Response logging for partial failures  is in 'DEBUG' level instead of 'INFO'
The table mentioned in -
https://github.com/googleads/google-ads-java#requestresponse-logging
is true only when exception is thrown, but not for partial failure response.

PR #121, opened 2019-05-24
Update logging instructions for partial failure levels

nwbirnie:
Fixes https://github.com/googleads/google-ads-java/issues/112

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,9 +263,8 @@ headers.
 | SUMMARY  | com.google.ads.googleads.lib.request.summary | INFO          | WARN          |
 | DETAIL   | com.google.ads.googleads.lib.request.detail  | DEBUG         | INFO          |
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -263,6 +263,10 @@ headers.
 | SUMMARY  | com.google.ads.googleads.lib.request.summary | INFO          | WARN          |
 | DETAIL   | com.google.ads.googleads.lib.request.detail  | DEBUG         | INFO          |
 
+**Caveat**: Partial failures does not cause the entire request to fail if some of its operations fail.
```
jradcliff: Readers may not know what partial failure is, so I suggest changing this to:

**Caveat**: Mutate requests where [`partialFailure`](https://developers.google.com/google-ads/api/reference/rpc/google.ads.googleads.v1.services) is `true` will not fail if individual operations fail.

PR #150, opened 2019-06-28
Prepare versions and change log for release 4.0.0

devchas:


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+4.0.0 - 2019-06-28
```
nwbirnie: I think this should be 3.1.0, which is a little counter intuitive, but since we're not breaking anything in the existing library there's no need for a major version bump...

devchas:
Created a new branch with correct name.

PR #162, opened 2019-07-18
Update examples to work with continuous integration.

nwbirnie:
- Make runExample() methods public.
- Take CodeSample params as parameter to runExamples()
- Fix a couple of flaky examples (e.g. depending on
System.currentTimeMillis() for unique Ids).

### DIFF
--- a/google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/CreateCustomer.java
+++ b/google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/CreateCustomer.java
@@ -33,15 +33,15 @@ import java.time.format.DateTimeFormatter;
 /**
  * Creates a new customer under a given manager account.
  *
- * Note: this example must be run using the credentials of a Google Ads manager account. By default,
- * the new account will only be accessible via the manager account.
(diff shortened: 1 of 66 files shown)


google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/CreateCustomer.java
```diff
@@ -83,32 +83,33 @@ public static void main(String[] args) {
     }
   }
 
-  private void runExample(GoogleAdsClient googleAdsClient, Long managerId) {
+  public void runExample(GoogleAdsClient googleAdsClient, CreateCustomerParams params) {
```
jradcliff: Back when we added command line params to AdWords API examples, we decided *against* passing the param object into `runExample`, since this would make it hard for users to copy/paste from examples without also copying over the param class.

nwbirnie: We can add an overload instead if that's better.

I'm not really convinced that this is necessary though - folks will need to rename the method anyway. I don't think changing the signature is asking much.

PR #199, opened 2019-12-16
Update README for enableGeneratedCatalog option

devchas:


README.md
```diff
@@ -193,6 +193,14 @@ the following format:
     # uses multiple manager account + OAuth pairs.
     #
     # api.googleads.loginCustomerId=INSERT_LOGIN_CUSTOEMR_ID_HERE
+    
+    # By default, this library uses reflection to build the ApiCatalog. 
+    # In order to reduce latency, users can set api.googleads.enableGeneratedCatalog=true,
```
jradcliff: Let's clarify where this will reduce latency, e.g.,

In order to reduce latency when creating service clients, ...

My main concern here is that I don't want to imply that this will change the latency of requests to the API, but instead is purely a client-side improvement during service client creation.

PR #211, opened 2019-12-23
Add comment clarifying empty result set

### LINKED ISSUE #210: Clarify empty result set in HandleRateExceededError.java
The result set of [HandleRateExceededError.java](https://github.com/googleads/google-ads-java/blob/9eb8459fab9cee843ef997038229220215a49eba/google-ads-examples/src/main/java/com/google/ads/googleads/examples/errorhandling/HandleRateExceededError.java#L211) will always be empty with the validateOnly flag set to true. We need to clarify this is expected behavior or print the number of results validated.

devchas:
Fixes #210

### DIFF
--- a/google-ads-examples/src/main/java/com/google/ads/googleads/examples/errorhandling/HandleRateExceededError.java
+++ b/google-ads-examples/src/main/java/com/google/ads/googleads/examples/errorhandling/HandleRateExceededError.java
@@ -209,12 +209,13 @@ public class HandleRateExceededError {
                       Long.toString(customerId), operations, false, true);
               System.out.printf("%d operations validated.%n", operations.size());
               // The result set will be empty while validateOnly is set to true in the
-              // mutateAdGroupCriteria request. Remove the validateOnly parameter or set to false
-              // to complete the request and return a result set.
-              System.out.printf("Added %d ad group criteria:%n", response.getResultsCount());
-              for (MutateAdGroupCriterionResult result : response.getResultsList()) {
-                System.out.println(result.getResourceName());
-              }
+              // mutateAdGroupCriteria request. Remove the validateOnly parameter (or set to false)
+              // and uncomment the code below to complete the request and return a result set.
+
+              //System.out.printf("Added %d ad group criteria:%n", response.getResultsCount());
+              //for (MutateAdGroupCriterionResult result : response.getResultsList()) {
+              //  System.out.println(result.getResourceName());
+              //}
               break;
             } catch (GoogleAdsException gae) {
               for (GoogleAdsError googleAdsError : gae.getGoogleAdsFailure().getErrorsList()) {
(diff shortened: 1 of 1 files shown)


google-ads-examples/src/main/java/com/google/ads/googleads/examples/errorhandling/HandleRateExceededError.java
```diff
@@ -207,7 +207,9 @@ public void run() {
               MutateAdGroupCriteriaResponse response =
                   adGroupCriterionServiceClient.mutateAdGroupCriteria(
                       Long.toString(customerId), operations, false, true);
-
+              // The result set will be empty while validateOnly is set to true in the
+              // mutateAdGroupCriteria request. Remove the validateOnly parameter or set to false
+              // to complete the request and return a result set.
```
AnashOommen: Does it make sense to do 
System.Out.PrintLn("%d operations validated.");

// The result set will be empty while validateOnly is set to true in the
// mutateAdGroupCriteria request. Remove the validateOnly parameter to false
// and uncomment the code below to complete the request and return a result set.
System.out.printf("Added %d ad group criteria:%n", response.getResultsCount());
for (MutateAdGroupCriterionResult result : response.getResultsList()) {
  System.out.println(result.getResourceName());

devchas: Yes, done.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/errorhandling/HandleRateExceededError.java
```diff
@@ -207,11 +207,15 @@ public void run() {
               MutateAdGroupCriteriaResponse response =
                   adGroupCriterionServiceClient.mutateAdGroupCriteria(
                       Long.toString(customerId), operations, false, true);
-
-              System.out.printf("Added %d ad group criteria:%n", response.getResultsCount());
-              for (MutateAdGroupCriterionResult result : response.getResultsList()) {
-                System.out.println(result.getResourceName());
-              }
+              System.out.printf("%d operations validated.%n", operations.size());
+              // The result set will be empty while validateOnly is set to true in the
```
jradcliff: These comments aren't easy to understand since we are not actually creating a MutateAdGroupCriteriaRequest, but instead are using the flattened method.
Since a key point of this example is the validateOnly parameter, I suggest creating a MutateAdGroupCriteriaRequest so readers can clearly see the call to the builder method for `validateOnly`.

sundquist: If we are trying to focus on demonstrating validate only, perhaps this commented out section should just be removed entirely? If we decide to keep it, perhaps, we can stick an 'if' block around it to execute or not based on if validate only is set or not.

devchas: I'm okay with removing the comment.

jradcliff: How about:
1. Switch to constructing a `MutateAdgroupCriteriaRequest` as mentioned above.
2. Making this comment simply:

> The result set will be empty since validateOnly is set to true in the MutateAdGroupCriteria request.

sundquist: +1 I like that idea.

devchas: See new commit.

PR #226, opened 2020-01-21
Add AddDemographicTargetingCriteria example

devchas:


google-ads-examples/src/main/java/com/google/ads/googleads/examples/targeting/AddDemographicTargetingCriteria.java
```diff
@@ -0,0 +1,138 @@
...
+   * @param googleAdsClient the Google Ads API client.
+   * @param customerId the client customer ID in which to create criterion.
+   * @param adGroupId the ad group ID in which to create criterion.
+   */
+  private void runExample(GoogleAdsClient googleAdsClient, long customerId, long adGroupId) {
+    String adGroupResourceName = ResourceNames.adGroup(customerId, adGroupId);
+
+    List<AdGroupCriterionOperation> operations = new ArrayList<>();
+
+    // Creates a gender demographic criteria.
```
jradcliff: s/criteria/criterion/ on this line and line 112.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/targeting/AddDemographicTargetingCriteria.java
```diff
@@ -0,0 +1,138 @@
...
+    AdGroupCriterion genderAdGroupCriterionOperation = AdGroupCriterion.newBuilder()
+        .setAdGroup(StringValue.of(adGroupResourceName))
+        .setGender(GenderInfo.newBuilder().setType(GenderType.MALE).build())
+        .build();
+
+    // Creates a age range negative demographic criteria.
```
jradcliff: nit: "a" -> "an"

google-ads-examples/src/main/java/com/google/ads/googleads/examples/targeting/AddDemographicTargetingCriteria.java
```diff
@@ -0,0 +1,138 @@
...
+   */
+  private void runExample(GoogleAdsClient googleAdsClient, long customerId, long adGroupId) {
+    String adGroupResourceName = ResourceNames.adGroup(customerId, adGroupId);
+
+    List<AdGroupCriterionOperation> operations = new ArrayList<>();
+
+    // Creates a gender demographic criteria.
+    AdGroupCriterion genderAdGroupCriterionOperation = AdGroupCriterion.newBuilder()
+        .setAdGroup(StringValue.of(adGroupResourceName))
+        .setGender(GenderInfo.newBuilder().setType(GenderType.MALE).build())
```
jradcliff: @fiboknacky I see this is based on the PHP example, which in turn was based on the AdWords API example. However, since demo criteria such as genders are a discrete set, adding a demo criteria without also setting a bid modifier or a bid doesn't serve much of a purpose. Setting a bid may be confusing because that brings in the whole question of [TargetRestriction](https://developers.google.com/google-ads/api/reference/rpc/google.ads.googleads.v2.common#google.ads.googleads.v2.common.TargetRestriction), but WDYT of having this at least apply a bid _modifier_?

devchas: Waiting on @fiboknacky response, but I've made that change for the time being.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/targeting/AddDemographicTargetingCriteria.java
```diff
@@ -98,22 +104,27 @@ public static void main(String[] args) {
    * @param customerId the client customer ID in which to create criterion.
    * @param adGroupId the ad group ID in which to create criterion.
```
jradcliff: Please add javadoc for the `bidModifier` param

google-ads-examples/src/main/java/com/google/ads/googleads/examples/targeting/AddDemographicTargetingCriteria.java
```diff
@@ -98,22 +104,27 @@ public static void main(String[] args) {
...
    */
-  private void runExample(GoogleAdsClient googleAdsClient, long customerId, long adGroupId) {
+  private void runExample(
+      GoogleAdsClient googleAdsClient,
+      long customerId,
+      long adGroupId,
+      Double bidModifier) {
```
jradcliff: Can make this a lowercase 'd' `double`, since you've made the parameter required

google-ads-examples/src/main/java/com/google/ads/googleads/examples/targeting/AddDemographicTargetingCriteria.java
fiboknacky: SGTM.

jradcliff: Very sorry for the back and forth here, but I got curious and tried out this code example, and what I found was that adding a positive `Gender` criterion at the ad group level implicitly excludes all other `Gender` criteria that were not explicitly added at the ad group level. The same applies to `AgeRange` and `ParentalStatus`.

In other words, the behavior is:
*  If you don't explicitly target or exclude any `Gender` criteria, all `Gender` criteria are implicitly targeted.
*  If you target any `Gender` criteria by adding an `AdGroupCriterion`, all other criteria you did not explicitly add are implicitly _excluded_.

In light of this, my suggestion to add a bid modifier doesn't necessarily make sense, as we're only targeting one `Gender`. Could you revert the addition I suggested?

Thanks

devchas: Done.

PR #228, opened 2020-01-24
Add UploadOfflineConversions example

devchas:
@AnashOommen because partial failure must be set to true for this example, I included code to print the partial failure if there is one as well as a check to see if the uploadConversion was successful before printing the final output. This wasn't present in the .NET example, but I think it is helpful and should be included.

### DIFF
--- a/google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadOfflineConversions.java
+++ b/google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadOfflineConversions.java
@@ -42,7 +42,7 @@ public class UploadOfflineConversions {
     @Parameter(names = ArgumentNames.CONVERSION_ACTION_ID, required = true)
     private long conversionActionId;
 
-    @Parameter(names = ArgumentNames.GCL_ID, required = true)
+    @Parameter(names = ArgumentNames.GCLID, required = true)
     private String gclid;
 
     @Parameter(names = ArgumentNames.CONVERSION_TIME, required = true,
--- a/google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadOfflineConversions.java
+++ b/google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadOfflineConversions.java
@@ -45,6 +45,9 @@ public class UploadOfflineConversions {
     @Parameter(names = ArgumentNames.GCL_ID, required = true)
     private String gclid;
 
+    // The date time at which the conversion occurred. Must be after the click time.
+    // The timezone must be specified.
+    // The format is "yyyy-mm-dd hh:mm:ss+|-hh:mm", e.g. “2019-01-01 12:32:45-08:00”.
     @Parameter(names = ArgumentNames.CONVERSION_TIME, required = true)
     private String conversionTime;
 
@@ -138,7 +141,9 @@ public class UploadOfflineConversions {
           .uploadClickConversions(
               Long.toString(customerId),
               ImmutableList.of(clickConversion),
+              // Enables partial failure (must be true).
               true,
+              // Disables validate only.
               false);
 
       // Prints any partial errors returned.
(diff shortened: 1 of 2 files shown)


google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadOfflineConversions.java
```diff
@@ -0,0 +1,162 @@
...
+
+    // Creates the conversion upload service client.
+    try (ConversionUploadServiceClient conversionUploadServiceClient =
+        googleAdsClient.getLatestVersion().createConversionUploadServiceClient()) {
+      // Uploads the click conversion. Partial failure should always be set to true.
+      UploadClickConversionsResponse response = conversionUploadServiceClient
+          .uploadClickConversions(
+              Long.toString(customerId),
+              ImmutableList.of(clickConversion),
+              true,
```
jradcliff: Please add comments here to indicate the parameters for the true and false values, .e.g,

```
   // Enable partial failure
   true,
   // Disable validate only
   false
```

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadOfflineConversions.java
```diff
@@ -0,0 +1,162 @@
...
+    @Parameter(names = ArgumentNames.CUSTOMER_ID, required = true)
+    private long customerId;
+
+    @Parameter(names = ArgumentNames.CONVERSION_ACTION_ID, required = true)
+    private long conversionActionId;
+
+    @Parameter(names = ArgumentNames.GCL_ID, required = true)
+    private String gclid;
+
+    @Parameter(names = ArgumentNames.CONVERSION_TIME, required = true)
```
jradcliff: Please add a description that explains the expected format of this parameter

jradcliff: Description looks good, but could you add it to the `description` attribute of the `@Parameter` annotation instead? That way, users will see it on the console when they run this example with `--help`.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadOfflineConversions.java
```diff
@@ -45,7 +45,10 @@
     @Parameter(names = ArgumentNames.GCL_ID, required = true)
     private String gclid;
 
-    @Parameter(names = ArgumentNames.CONVERSION_TIME, required = true)
+    @Parameter(names = ArgumentNames.CONVERSION_TIME, required = true,
+        description = "The date time at which the conversion occurred. "
+            + "Must be after the click time. "
+            + "The format is  \"yyyy-mm-dd hh:mm:ss+|-hh:mm\", e.g. “2019-01-01 12:32:45-08:00”.")
```
jradcliff: There are some slanted quotes in here. How about using single quotes within the string instead?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadOfflineConversions.java
```diff
@@ -45,7 +45,10 @@
     @Parameter(names = ArgumentNames.GCL_ID, required = true)
     private String gclid;
 
-    @Parameter(names = ArgumentNames.CONVERSION_TIME, required = true)
+    @Parameter(names = ArgumentNames.CONVERSION_TIME, required = true,
+        description = "The date time at which the conversion occurred. "
+            + "Must be after the click time. "
```
jradcliff: "Must be after the click time, and must include the time zone offset."

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadOfflineConversions.java
```diff
@@ -0,0 +1,167 @@
...
+public class UploadOfflineConversions {
+  private static class UploadOfflineConversionsParams extends CodeSampleParams {
+
+    @Parameter(names = ArgumentNames.CUSTOMER_ID, required = true)
+    private long customerId;
+
+    @Parameter(names = ArgumentNames.CONVERSION_ACTION_ID, required = true)
+    private long conversionActionId;
+
+    @Parameter(names = ArgumentNames.GCL_ID, required = true)
```
AnashOommen: GCLID is used as a single word.

PR #229, opened 2020-01-24
Add UploadImageAsset example

devchas:


google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/UploadImageAsset.java
```diff
@@ -0,0 +1,130 @@
...
+    try (AssetServiceClient assetServiceClient =
+        googleAdsClient.getLatestVersion().createAssetServiceClient()) {
+      // Issues a mutate request to add the asset.
+      MutateAssetsResponse response =
+          assetServiceClient.mutateAssets(Long.toString(customerId), ImmutableList.of(operation));
+      if (!response.getResultsList().isEmpty()) {
+        System.out.printf("The image asset with resource name '%s' was created.%n",
+            response.getResults(0).getResourceName());
+      } else {
+        System.out.print("No image asset was created.%n");
```
nwbirnie: Why would this happen? Surely if the asset wasn't created the RPC would fail...

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/UploadImageAsset.java
```diff
@@ -0,0 +1,130 @@
...
+
+  /**
+   * Runs the example.
+   *
+   * @param googleAdsClient the Google Ads API client.
+   * @param customerId the client customer ID.
+   * @throws IOException if there are errors related to image processing.
+   */
+  private void runExample(GoogleAdsClient googleAdsClient, long customerId) throws IOException {
+
```
nwbirnie: nit: remove blank line

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/UploadImageAsset.java
devchas: I had followed the PHP example but agree with you. Perhaps the behavior is different in PHP. I'll change that.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/UploadImageAsset.java
```diff
@@ -0,0 +1,126 @@
...
+        .setData(BytesValue.of(ByteString.copyFrom(imageData)))
+        .build();
+
+    // Creates an asset.
+    Asset asset = Asset.newBuilder()
+        // Optional: Provide a unique friendly name to identify your asset.
+        // If you specify the name field, then both the asset name and the image being
+        // uploaded should be unique, and should not match another ACTIVE asset in this
+        // customer account.
+        // .setName(StringValue.of("Jupiter Trip # " +  + System.currentTimeMillis()))
```
PierrickVoulet: There are extra "+" and space characters

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/UploadImageAsset.java
```diff
@@ -0,0 +1,126 @@
...
+    AssetOperation operation =  AssetOperation.newBuilder().setCreate(asset).build();
+
+    // Creates the service client.
+    try (AssetServiceClient assetServiceClient =
+        googleAdsClient.getLatestVersion().createAssetServiceClient()) {
+      // Issues a mutate request to add the asset.
+      MutateAssetsResponse response =
+          assetServiceClient.mutateAssets(Long.toString(customerId), ImmutableList.of(operation));
+      // Prints the result.
+      System.out.printf("The image asset with resource name '%s' was created.%n",
```
PierrickVoulet: nit: it should theoretically never happen but you might want to add an extra condition to avoid out-of-bounds exception if nothing is returned in the results.

nwbirnie: I don't think we should cover this case - the expectation should be a successful RPC for X operations returns X results.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/UploadImageAsset.java
nwbirnie: Can you flag this with the PHP owners? I don't think this is library specific - I'm guessing this is unreachable code.

devchas: @PierrickVoulet ^^

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/UploadImageAsset.java
devchas: Good catch, thx!

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/UploadImageAsset.java
PierrickVoulet: I think it is nit. We will keep it as-is for now in PHP to stay consistent with other code examples.

PR #238, opened 2020-02-10
Update CONTRIBUTING.MD

devchas:


CONTRIBUTING.md
```diff
@@ -1,28 +1,124 @@
...
 
-You generally only need to submit a CLA once, so if you've already submitted one
-(even if it was for a different project), you probably don't need to do it
-again.
+## Core API vs. Language Library
 
-## Code reviews
+This repository is specifically for the Google Ads API Client Library for Java. If you have feedback
+related to the core API, please submit it
+[here](https://developers.google.com/google-ads/api/support).
```
jradcliff: Should we point them to the **Send feedback** link on that page specifically? That will allow users to choose between providing documentation or API feedback.

CONTRIBUTING.md
```diff
@@ -1,28 +1,124 @@
...
+ * [Community Guidelines](#community-guidelines)
 
-You generally only need to submit a CLA once, so if you've already submitted one
-(even if it was for a different project), you probably don't need to do it
-again.
+## Core API vs. Language Library
 
-## Code reviews
+This repository is specifically for the Google Ads API Client Library for Java. If you have feedback
+related to the core API, please submit it
```
jradcliff: "core API" may be unclear. Perhaps "the Google Ads API" or just "the API" instead?

CONTRIBUTING.md
```diff
@@ -1,28 +1,124 @@
...
+    *  [Pull Request Submission Best Practices](#pull-request-submission-best-practices)
+    *  [Style Guidelines & Naming Conventions](#style-guidelines--naming-conventions)
+    *  [Claiming Requests](#claiming-requests)
+    *  [Testing](#testing)
+ * [Community Guidelines](#community-guidelines)
 
-You generally only need to submit a CLA once, so if you've already submitted one
-(even if it was for a different project), you probably don't need to do it
-again.
+## Core API vs. Language Library
```
jradcliff: See comment below

jradcliff: s/Language Library/Library for Java/

jradcliff: I don't think we need a separate **Core API vs....** section. The paragraph about where to file API feedback can just be the first paragraph under the **Providing Feedback** heading.

CONTRIBUTING.md
```diff
@@ -1,28 +1,121 @@
...
 
-All submissions, including submissions by project members, require review. We
-use GitHub pull requests for this purpose. Consult
-[GitHub Help](https://help.github.com/articles/about-pull-requests/) for more
-information on using pull requests.
+### Feature Requests
+
+If you see that someone has already created an issue that describes your proposed feature request,
+please upvote their idea to avoid duplicates and help our team prioritize feature requests.
+Github even provides a similar issues search as you create new issues, which can help in this
```
nwbirnie: I feel like the word "even" reads like "Github will do this for you, so don't be lazy". Suggest we drop "even" from this sentence.

CONTRIBUTING.md
```diff
@@ -1,28 +1,121 @@
...
+This repository is specifically for the Google Ads API Client Library for Java. If you have feedback
+related to the Google Ads API, please submit it by clicking the "Send feedback" button in the top right
+corner of [this page](https://developers.google.com/google-ads/api/support).
 
-All submissions, including submissions by project members, require review. We
-use GitHub pull requests for this purpose. Consult
-[GitHub Help](https://help.github.com/articles/about-pull-requests/) for more
-information on using pull requests.
+### Feature Requests
+
```
nwbirnie: Would be good to start FR + Bugs sections with some positive encouragement - currently the first thing in these sections is a prescriptive list of steps to take *before* submitting a ticket. We love getting feedback, and if folks read nothing else, I would like it to be telling them to go ahead and report the issue, even if that means we get more issues in the inbox. I'm happy to triage these requests as they come in and mark dupes etc.

devchas: The opening paragraph contains that language. Do you think we need it here as well?

CONTRIBUTING.md
```diff
@@ -21,19 +20,17 @@ contributing, please review the guidelines and best practices below.
...
 
-## Core API vs. Language Library
+## Providing Feedback
 
 This repository is specifically for the Google Ads API Client Library for Java. If you have feedback
-related to the core API, please submit it
-[here](https://developers.google.com/google-ads/api/support).
-
-## Providing Feedback
+related to the Google Ads API, please submit it by clicking the "Send feedback" button in the top right
```
jradcliff: nit: Put **Send feedback** in bold. This is the [standard on developers.google.com for any UI elements](https://developers.google.com/style/ui-elements#formatting).

PR #239, opened 2020-02-10
Add ValidateTextAd example

devchas:


google-ads-examples/src/main/java/com/google/ads/googleads/examples/campaignmanagement/ValidateTextAd.java
```diff
@@ -0,0 +1,177 @@
...
+   */
+  private void runExample(GoogleAdsClient googleAdsClient, long customerId, long adGroupId) {
+    String adGroupResourceName = ResourceNames.adGroup(customerId, adGroupId);
+
+    // Creates the expanded text ad info.
+    ExpandedTextAdInfo expandedTextAdInfo =
+        ExpandedTextAdInfo.newBuilder()
+            .setDescription(StringValue.of("Luxury Cruise to Mars"))
+            .setHeadlinePart1(StringValue.of("Visit the Red Planet in style"))
+            .setHeadlinePart2(StringValue.of("Low-gravity fun for everyone!!"))
```
AnashOommen: Nit: You should probably call out here that this is a policy violation that has been introduced on purpose to illustrate the error handling. Add a row on the prio sheet to let others know that this needs a fix.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/campaignmanagement/ValidateTextAd.java
```diff
@@ -0,0 +1,178 @@
...
+    AdGroupAdOperation operation = AdGroupAdOperation.newBuilder().setCreate(adGroupAd).build();
+
+    try (AdGroupAdServiceClient adGroupAdServiceClient =
+        googleAdsClient.getLatestVersion().createAdGroupAdServiceClient()) {
+      // Issues the mutate request setting validateOnly=true.
+      MutateAdGroupAdsResponse response =
+          adGroupAdServiceClient.mutateAdGroupAds(
+              Long.toString(customerId), ImmutableList.of(operation), false, true);
+
+      // Since validation is ON, result will be null.
```
nwbirnie: Is this code effectively unreachable? If so should we not flag this as an error, e.g. "Expanded text ad passed validation, this is unexpected."

devchas: This code is unreachable in the current state. However, if you were to remove the policy error in the headline, the sample will print "Expanded text ad validated successfully." which I think is the appropriate behavior.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/campaignmanagement/ValidateTextAd.java
```diff
@@ -0,0 +1,178 @@
...
+import com.google.ads.googleads.v2.utils.ResourceNames;
+import com.google.common.collect.ImmutableList;
+import com.google.protobuf.StringValue;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Shows how to use the validateOnly header to validate an expanded text ad. No objects will be
```
nwbirnie: I think "no ads entities will be created" is a little more accurate here, there will technically be several objects created in the process, the changes won't be persisted though.

PR #241, opened 2020-02-20
Add AddUnionUserList example

devchas:
This example is for the Google Ads API remarketing guide.


google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddListBasedUserList.java
```diff
@@ -0,0 +1,152 @@
...
+            .setOperator(UserListLogicalRuleOperator.ANY)
+            .addAllRuleOperands(logicalUserListOperandInfoList)
+            .build();
+
+    // Creates the new combination user list.
+    UserList userList =
+        UserList.newBuilder()
+            .setName(
+                StringValue.of(
+                    "My combination list of Mars or Venus customers #"
```
AnashOommen: There should be some additional comments to set the context here. In this code example, we are accepting a list of userListIDs, and it is not very clear where the reference to Mars customers or Venus customers come from. I would have expected the runExample method to accept a marsUserListId and venusUserListId in this case, for example.

devchas: I updated the code example after a first draft to accept a list of userListIDs rather than two specific ones and forgot to update the name of the new combined list. I'll update that name. Thanks for catching this.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddListBasedUserList.java
```diff
@@ -0,0 +1,152 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Creates a combination user list containing users that are present on any one of the provided user
+ * lists.
+ */
+public class AddListBasedUserList {
```
nwbirnie: Could we have a more descriptive name for this example perhaps? I find AddListBasedUserList confusing, partly due to the repitition of "list". How about AddUnionUserList, AddCombinationUserList, MergeUserLists?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddListBasedUserList.java
```diff
@@ -0,0 +1,152 @@
...
+   * Runs the example.
+   *
+   * @param googleAdsClient the Google Ads API client.
+   * @param customerId the client customer ID.
+   * @param userListIds the IDs of the lists to be used for the new combination user list.
+   * @throws GoogleAdsException if an API request failed with one or more service errors.
+   */
+  private void runExample(
+      GoogleAdsClient googleAdsClient, long customerId, List<Long> userListIds) {
+    // Creates the rule operand info targeting the provided list IDs.
```
nwbirnie: Not sure I understand this comment - could you describe why this is being done rather than what is being done?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddListBasedUserList.java
```diff
@@ -0,0 +1,152 @@
...
+          LogicalUserListOperandInfo.newBuilder()
+              .setUserList(StringValue.of(userListResourceName))
+              .build());
+    }
+
+    // Creates the UserListLogicalRuleInfo specifying that a user should be added to the new list if
+    // they are present in any of the provided lists.
+    UserListLogicalRuleInfo userListLogicalRuleInfo =
+        UserListLogicalRuleInfo.newBuilder()
+            .setOperator(UserListLogicalRuleOperator.ANY)
```
nwbirnie: How about adding some optional parenthesis here, to illustrate the various different operators? This can be quite a powerful tool when used well, would be nice to showcase this a little.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddListBasedUserList.java
nwbirnie: Another option might be AddLogicalUserList if we don't expect any other examples with the same name.

devchas: I like AddUnionUser list. Given that there are several other user list examples, I want to be specific about what this example does as to avoid confusion.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddListBasedUserList.java
devchas: I thought it might be confusing to list conflicting operators (even if commented out), so I have added a detailed comment explaining the options instead.

PR #268, opened 2020-03-30
Update dependency versions for consistency

jradcliff:
* Move to `netty-tcnative-boringssl-static` version `2.0.26.Final` and protobuf version `3.10.0` to match to match target gRPC version `1.25.0`.
* Move more recent versions of `auto-value` and `protoc-jar-maven-plugin`.
* Make log4j dependencies in examples modules have `scope` of `runtime` since they are not needed for compilation.

### DIFF
--- a/google-ads/pom.xml
+++ b/google-ads/pom.xml
@@ -33,10 +33,10 @@
   </description>
 
   <properties>
-    <auto-value.version>1.4</auto-value.version>
+    <auto-value.version>1.6.6</auto-value.version>
     <gax.version>1.50.1</gax.version>
     <grpc.version>1.25.0</grpc.version>
-    <protobuf.version>3.5.1</protobuf.version>
+    <protobuf.version>3.10.0</protobuf.version>
   </properties>
 
   <dependencies>
@@ -82,7 +82,7 @@
       <!--
         See https://github.com/grpc/grpc-java/blob/master/SECURITY.md#troubleshooting
       -->
-      <version>2.0.7.Final</version>
+      <version>2.0.26.Final</version>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
@@ -136,7 +136,7 @@
       <plugin>
         <groupId>com.github.os72</groupId>
         <artifactId>protoc-jar-maven-plugin</artifactId>
-        <version>3.7.0</version>
+        <version>3.11.1</version>
         <executions>
           <execution>
             <id>generate-test-protos</id>
@@ -145,7 +145,7 @@
               <goal>run</goal>
             </goals>
             <configuration>
-              <protocArtifact>com.google.protobuf:protoc:3.7.0</protocArtifact>
+              <protocArtifact>com.google.protobuf:protoc:${protobuf.version}</protocArtifact>
               <inputDirectories>
                 <include>src/test/resources/protos</include>
               </inputDirectories>
--- a/google-ads/pom.xml
+++ b/google-ads/pom.xml
@@ -34,6 +34,8 @@
 
   <properties>
     <auto-value.version>1.6.6</auto-value.version>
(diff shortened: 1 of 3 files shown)


jradcliff:
Please see the table in the [**Netty** section here](https://github.com/grpc/grpc-java/blob/master/SECURITY.md#netty) for gRPC + Netty version compatibility.

jradcliff:
Friendly ping for @nwbirnie and @devchas for review

google-ads/pom.xml
```diff
@@ -33,10 +33,10 @@
   </description>
 
   <properties>
-    <auto-value.version>1.4</auto-value.version>
+    <auto-value.version>1.6.6</auto-value.version>
     <gax.version>1.50.1</gax.version>
     <grpc.version>1.25.0</grpc.version>
-    <protobuf.version>3.5.1</protobuf.version>
+    <protobuf.version>3.10.0</protobuf.version>
```
nwbirnie, line 39: Can we remove these dependencies and defer to the version requested by GAX?

jradcliff: _Almost_, with the exception of the transitive dependency in tests on `grpc-context`. See additional `test` scope dependency below.

I'm keeping these properties because putting them all side-by-side with the netty version will make it easier to verify consistency going forward.

google-ads/pom.xml
```diff
@@ -116,6 +98,17 @@
       <version>${gax.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <!--
+      Without this dependency, gax will add io.grpc:grpc-context:jar:1.22.1 to
```
nwbirnie, line 103: Doesn't this mean that we've got the wrong version of GAX? If the usage of `Deadline` is in generated test code, presumably it was generated for some version of GAX that uses gRPC 1.25.0.

jradcliff: Not necessarily. Here's a snippet of the `dependency:tree` output _without_ this change:

```
[INFO] com.google.api-ads:google-ads:jar:5.0.1-SNAPSHOT
[INFO] +- com.google.api:gax:jar:1.50.1:compile
...
[INFO] |  +- com.google.api:api-common:jar:1.8.1:compile
[INFO] |  \- io.opencensus:opencensus-api:jar:0.24.0:compile
[INFO] |     \- io.grpc:grpc-context:jar:1.22.1:compile  <-- ---- The ONE grpc dep on 1.22.1
[INFO] +- com.google.api:gax-grpc:jar:1.50.1:compile
[INFO] |  +- io.grpc:grpc-stub:jar:1.25.0:compile        <-- ++++ Everything under gax-grpc 1.50.1 uses 1.25.0
[INFO] |  |  \- io.grpc:grpc-api:jar:1.25.0:compile
[INFO] |  +- io.grpc:grpc-auth:jar:1.25.0:compile
[INFO] |  +- io.grpc:grpc-protobuf:jar:1.25.0:compile
[INFO] |  |  +- com.google.protobuf:protobuf-java:jar:3.10.0:compile
[INFO] |  |  \- io.grpc:grpc-protobuf-lite:jar:1.25.0:compile
```
Notice how:
1. `1.22.1` of `grpc-context` is included as a transitive dependency of `gax 1.50.1 -> api-common -> opencensus-api`
1. All other `grpc` dependencies via `gax-grpc 1.50.1` are at `1.25.0`

It seems to me that the inconsistency is in `gax 1.50.1`, as the `grpc` version it's using indirectly for this one grpc package is inconsistent with the `grpc` version it's using for everything else.

nwbirnie: Oh, I see. Thanks for checking dude.

chuckbartz26:
[]()da004ba646af2a580f34c391c0b8847f45d05030@jradcliff .com

PR #269, opened 2020-04-03
Add GetAccountHierarchy example

devchas:


google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,260 @@
...
+                + "customer_client.manager, customer_client.descriptive_name, "
+                + "customer_client.currency_code, customer_client.time_zone, "
+                + "customer_client.id "
+                + "FROM customer_client "
+                + "WHERE customer_client.level <= 1";
+
+        // Performs a breadth-first search algorithm to build a mapping of managers to their
+        // child accounts. The map keys are manager IDs, and the values are lists of those
+        // managers' direct child accounts, as represented by CustomerClient objects.
+        while (!unprocessedCustomerIds.isEmpty()) {
```
nwbirnie, line 151: This method seems to be getting quite long and hard to follow - there's multiple levels of loop nesting, data structures, API calls and a reasonably complicated algorithm wrapping the whole thing together. Would suggest refactoring with extract method to make the code more self-documenting.

Tip: the amount and complexity of comments can be a [code smell](https://refactoring.guru/smells/comments).

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,260 @@
...
+          googleAdsClient.getLatestVersion().createGoogleAdsServiceClient()) {
+        // Creates a query that retrieves all child accounts of the manager specified in search
+        // calls below.
+        String query =
+            "SELECT customer_client.client_customer, customer_client.level, "
+                + "customer_client.manager, customer_client.descriptive_name, "
+                + "customer_client.currency_code, customer_client.time_zone, "
+                + "customer_client.id "
+                + "FROM customer_client "
+                + "WHERE customer_client.level <= 1";
```
nwbirnie, line 146: Would it not be simpler to just pull the entire hierarchy at once and convert the result to a map? Am I missing something here that requires us to implement BFS? You can get the parent customer ID from `CustomerName.parse(resourceName).getCustomerId()`.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,260 @@
...
+    for (long seedCustomerId : seedCustomerIds) {
+      // Instantiates several variables used later.
+      Map<Long, List<CustomerClient>> customerIdsToChildAccounts = new HashMap<>();
+      CustomerClient rootCustomerClient = null;
+      int depth = 0;
+      // Adds the seed customer ID to the list of IDs to be processed.
+      unprocessedCustomerIds.add(seedCustomerId);
+
+      // Creates the Google Ads Service client.
+      try (GoogleAdsServiceClient googleAdsServiceClient =
```
nwbirnie, line 136: Do we need to change the `loginCustomerId` here? Say for instance `listAccessibleCustomers()` returns 3 different manager accounts, we'll need to set `loginCustomerId` differently for each call to `select client_customer`.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,260 @@
...
+                + "customer_client.id "
+                + "FROM customer_client "
+                + "WHERE customer_client.level <= 1";
+
+        // Performs a breadth-first search algorithm to build a mapping of managers to their
+        // child accounts. The map keys are manager IDs, and the values are lists of those
+        // managers' direct child accounts, as represented by CustomerClient objects.
+        while (!unprocessedCustomerIds.isEmpty()) {
+          long customerId = unprocessedCustomerIds.get(0);
+          unprocessedCustomerIds.remove(0);
```
nwbirnie, line 153: Can do this in one line. `List.remove(int index)` will return the removed object.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,260 @@
...
+          // Issues a search request by specifying page size.
```
nwbirnie, line 154: I'll skip reviewing below here while we resolve the comments above.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,260 @@
...
+    }
+  }
+
+  /**
+   * Prints the specified account's hierarchy using recursion.
+   *
+   * @param customerClient the customer client whose info will be printed and its child accounts
+   *     will be processed if it's a manager.
+   * @param customerIdsToChildAccounts a map containing the account hierarchy information. The keys
+   *     are manager account IDs, and the values are lists ofthe children for given manager
```
fiboknacky, line 228: Need a space between "of" and "the"

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,260 @@
...
+          System.out.println(customer);
+          seedCustomerIds.add(customer);
+        }
+      }
+    }
+
+    List<Long> unprocessedCustomerIds = new ArrayList<>();
+    for (long seedCustomerId : seedCustomerIds) {
+      // Instantiates several variables used later.
+      Map<Long, List<CustomerClient>> customerIdsToChildAccounts = new HashMap<>();
```
fiboknacky, line 129: This is recommended by Google to use guava's ListMultiMap

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,260 @@
...
+            // Updates the customerClientChildren with the list of children accounts if there
+            // exists an entry in the map for the given manager.
+            if (customerIdsToChildAccounts.containsKey(customerId)) {
+              customerClientChildren = customerIdsToChildAccounts.get(customerId);
+            }
+            // Adds the current child account to the list of the given manager's child accounts,
+            // which will be empty if the manager has not yet been added to the map.
+            customerClientChildren.add(customerClient);
+            // Inserts or updates the map for the given manager with the new list of child accounts.
+            customerIdsToChildAccounts.put(customerId, customerClientChildren);
```
fiboknacky, line 193: Overall, I'm fine with a large number of comments in other places of this file.
However, for these lines, I found it's too bulky and rather make the code hard to read.

Please consider using ListMultiMap. I believe it would make this part much simpler.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
fiboknacky: Chiming in quickly: I believe we cannot pull the entire hierarchy. There were some edge cases that we found back then, when implementing the golden example.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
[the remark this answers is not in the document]
fiboknacky: Oh. I'm too slow for this. :-P
+1 We should use Guava here.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,260 @@
...
+            // Checks is the child account is a manager itself so that it can later be processed
+            // and added to the map if it hasn't been already.
+            if (customerClient.getManager().getValue()) {
+              // A customer can be managed by multiple managers, so to prevent visiting the same
+              // customer multiple times, we need to check if it's already in the map.
+              boolean alreadyVisited =
+                  customerIdsToChildAccounts.containsKey(customerClient.getId().getValue());
+              if (!alreadyVisited && customerClient.getLevel().getValue() == 1) {
+                unprocessedCustomerIds.add(customerClient.getId().getValue());
```
nwbirnie, line 202: Can make this list immutable by adding all manager accounts from the hierarchy (as retrieved in the first API call with the query update I suggested above). That might help make the code easier to understand & maintain.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,260 @@
...
+          }
+        }
+
+        if (!(rootCustomerClient == null)) {
+          System.out.printf(
+              "The hierarchy of customer ID %d is printed below.%n",
+              rootCustomerClient.getId().getValue());
+          printAccountHierarchy(rootCustomerClient, customerIdsToChildAccounts, depth);
+        } else {
+          System.out.println(
```
nwbirnie, line 214: I thought this was fixed in [b/141760006](https://b/141760006)? If this is still happening it's a bug that should be reported and fixed.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,260 @@
...
+    }
+  }
+
+  /**
+   * Runs the examples.
+   *
+   * @param googleAdsClient the Google Ads API client.
+   * @param managerId managerCustomerId the customer ID.
+   */
+  private void runExample(GoogleAdsClient googleAdsClient, Long managerId) {
```
nwbirnie, line 102: Would suggest testing this on a large and complex manager hierarchy.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,260 @@
...
+          googleAdsClient.getLatestVersion().createCustomerServiceClient()) {
+        ListAccessibleCustomersResponse accessibleCustomers =
+            customerServiceClient.listAccessibleCustomers(
+                ListAccessibleCustomersRequest.newBuilder().build());
+        System.out.println(
+            "No manager customer ID is specified. The example will print the "
+                + "hierarchies of all accessible customer IDs:");
+
+        for (String customerResourceName : accessibleCustomers.getResourceNamesList()) {
+          long customer =
```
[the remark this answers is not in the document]
devchas: Yes, that is google-java-format.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
devchas: unprocessedCustomerIds (now named managerAccountsToSearch) has been changed to a set, so this has changed and it now a moot point.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,260 @@
...
+        while (!unprocessedCustomerIds.isEmpty()) {
+          long customerId = unprocessedCustomerIds.get(0);
+          unprocessedCustomerIds.remove(0);
+          // Issues a search request by specifying page size.
+          SearchPagedResponse response =
+              googleAdsServiceClient.search(
+                  SearchGoogleAdsRequest.newBuilder()
+                      .setQuery(query)
+                      .setCustomerId(Long.toString(customerId))
+                      .setPageSize(PAGE_SIZE)
```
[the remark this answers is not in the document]
devchas: Removing here.

@fiboknacky - FYI this still exists in PHP.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,260 @@
...
+                      .build());
+
+          // Iterates over all rows in all pages to get all customer clients under the specified
+          // customer's hierarchy.
+          for (GoogleAdsRow googleAdsRow : response.iterateAll()) {
+            CustomerClient customerClient = googleAdsRow.getCustomerClient();
+            // The customer client making the request will be returned in the
+            // search result with a level of 0.
+            if (customerClient.getLevel().getValue() == 0) {
+              // Store the root customer client, which is the first encountered customer client
```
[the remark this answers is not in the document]
devchas: This is correct is as is. rootCustomerClient is initialized as null. Upon encountering the first level 0 customerClient, we store that value as the rootCustomerClient. In all subsequent circumstances when the level is 0, rootCustomerClient will no longer = null, so the if statement is false, and we do not reassign.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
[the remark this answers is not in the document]
devchas: Switched to MultiMap 👍

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
devchas: I have test against a hierarchy with multiple levels of managers containing a mix of clients and managers in each level.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
devchas: I believe making the managerId param required makes this a moot point.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
devchas: Should be simpler now. WDYT?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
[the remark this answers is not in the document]
fiboknacky: How is setting page size is an anti-pattern?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -18,50 +18,37 @@
...
- */
+/** Gets the account hierarchy of the specified manager account. */
 public class GetAccountHierarchy {
 
-  private static final int PAGE_SIZE = 1_000;
-
   private static class GetAccountHierarchyParams extends CodeSampleParams {
 
-    @Parameter(names = ArgumentNames.MANAGER_ID)
+    @Parameter(names = ArgumentNames.MANAGER_ID, required = true)
```
fiboknacky, line 41: Could you ensure we agree on changing this behavior first?
It was suggested by two reviewers when creating PHP golden examples, and it was implemented already in Perl.
This will make this library diverge from other libraries.

devchas: Change reverted.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,296 @@
...
+            // Iterates over all rows in all pages to get all customer clients under the specified
+            // customer's hierarchy.
+            for (GoogleAdsRow googleAdsRow : response.iterateAll()) {
+              CustomerClient customerClient = googleAdsRow.getCustomerClient();
+              // The customer client making the request will be returned in the
+              // search result with a level of 0.
+              if (customerClient.getLevel().getValue() == 0) {
+                // Store the root customer client, which is the first encountered customer client
+                // that has level of 0.
+                if (rootCustomerClient == null) {
```
nwbirnie, line 217: Can you explain why we need to get this from the API? Can't we just store the first customer ID we requested the customer clients for?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,296 @@
...
+
+        for (String customerResourceName : accessibleCustomers.getResourceNamesList()) {
+          long customer = Long.parseLong(CustomerName.parse(customerResourceName).getCustomer());
+          System.out.println(customer);
+          seedCustomerIds.add(customer);
+        }
+      }
+    }
+
+    for (long seedCustomerId : seedCustomerIds) {
```
nwbirnie, line 162: Can we extract a method for the body of this loop. Have that return a Map. Then for each iteration of the outer loop, merge the map returned with the previous?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,296 @@
...
+   * @param googleAdsClient the Google Ads API client.
+   * @param managerId the root customer ID from which to begin the search.
+   */
+  private void runExample(GoogleAdsClient googleAdsClient, Long managerId, Long loginCustomerId) {
+    List<Long> seedCustomerIds = new ArrayList<>();
+    if (managerId != null) {
+      seedCustomerIds.add(managerId);
+    } else {
+      // Issues a request for listing all accessible customers by this authenticated Google account.
+      try (CustomerServiceClient customerServiceClient =
```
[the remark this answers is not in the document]
devchas: I do below, so I actually don't think this is needed here at all. To remove.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
devchas: We technically don't need this to find the root customer ID. This will simply be the seedCustomerId. We do, however, need two checks. We need to check if the current customer is the local root, in which case we'll continue - that's the if statement outside of this one. We also need to check if the current customer is the global root (or seed ID) so that we can save the customerClient to a local variable because we need the full object (not just the ID) for the printAccountHierarchy method. This would be the first time we'd have access to the full object. If it's provided as a parameter, it's only provided as an ID. If retrieved from listAccessibleCustomers, we'll still only have the resources names.

We can refactor the nested if statements as follows to make it clearer.

```
              if (customerClient.getId().getValue() == seedCustomerId) {
                rootCustomerClient = customerClient;
              }

              // The steps below map parent and children accounts. Continue here so that managers
              // accounts exclude themselves from the list of their children accounts.
              if (customerClient.getId().getValue() == customerId) {
                continue;
              }
```

@fiboknacky you wrote this in the PHP golden, so please let me know if I've missed anything or if you have any input.

devchas: Also @nwbirnie - not sure if this was your question, but the results from the search query are not guaranteed to come back in order, so the local root may not be the first result.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -105,67 +226,99 @@ private void runExample(GoogleAdsClient googleAdsClient, long managerId) {
...
-            // customer multiple times, we need to check if it's already in the map.
-            boolean alreadyVisited =
-                customerIdsToChildAccounts.containsKey(customerClient.getId().getValue());
-            if (!alreadyVisited && customerClient.getLevel().getValue() == 1) {
-              managerAccountsToSearch.add(customerClient.getId().getValue());
+            // For all level-1 (direct child) accounts that are manager accounts, the above
+            // query will be run against them to create a map of managers to their
+            // child accounts for printing the hierarchy afterwards.
+            customerIdsToChildAccounts.put(customerId, customerClient);
+            // Checks is the child account is a manager itself so that it can later be processed
```
fiboknacky, line 265: "Checks if" not "Checks is"

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -105,67 +226,99 @@ private void runExample(GoogleAdsClient googleAdsClient, long managerId) {
...
-      if (!(rootCustomerClient == null)) {
-        System.out.printf(
-            "The hierarchy of customer ID %d is printed below.%n",
-            rootCustomerClient.getId().getValue());
-        printAccountHierarchy(rootCustomerClient, customerIdsToChildAccounts, depth);
-      } else {
-        System.out.println(
-            "Customer ID is likely a test account, so its customer client "
-                + "information cannot be retrieved");
+      if (rootCustomerClient == null) {
```
fiboknacky, line 284: Could you also include brief comment here what could cause the root customer client to be null?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -105,67 +226,99 @@ private void runExample(GoogleAdsClient googleAdsClient, long managerId) {
...
+        return null;
+      }
+
+      CustomerClient finalRootCustomerClient = rootCustomerClient;
```
fiboknacky, line 288: Why do we need a new variable here?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -80,16 +107,110 @@ public static void main(String[] args) {
...
+    } else {
+      seedCustomerIds = getAccessibleCustomers(googleAdsClient);
+    }
+
+    Map<CustomerClient, Multimap<Long, CustomerClient>> customerClientsToHierarchies =
+        new HashMap<>();
+    List<Long> accountsWithNoInfo = new ArrayList<>();
+    // Constructs a map of account hierarchies.
+    for (long seedCustomerId : seedCustomerIds) {
+      Map<CustomerClient, Multimap<Long, CustomerClient>> customerClientToHierarchy =
```
fiboknacky, line 155: There are customerClientToHierarchy and customerClientsToHiearchies.
This is quite confusing, considering that the type of the two variables are actually the same.
Could you name them in a different way?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -80,16 +107,110 @@ public static void main(String[] args) {
...
+          createCustomerClientToHierarchy(loginCustomerId, seedCustomerId);
```
fiboknacky, line 156: The createCustomerClientToHierarchy method should actually be named as createCustomerClientsToHierarchies.
They contain many customer clients and many hierarchies.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
fiboknacky: I'd, e.g., simply name the latter as "allHierarchies".

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -21,34 +21,60 @@
...
+      // example will use each respective accessible customer ID as the loginCustomerId.
+      // params.managerId = Long.parseLong("INSERT_MANAGER_ID_HERE");
+    }
+
+    if (params.managerId != null && params.loginCustomerId == null) {
+      System.err.println("loginCustomerId must be provided if managerId is provided.");
+      return;
+    } else if (params.managerId == null && params.loginCustomerId != null) {
+      System.err.println("loginCustomerId may not be provided if managerId is not provided.");
+      return;
```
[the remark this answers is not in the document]
devchas: That would make this more concise, but we lose the specificity in the error messages, which I think will be helpful in practice.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
devchas: While the return value is technically a map of CustomerClient keys and Multimap values, it always returns a map with a single key, value pair, which is why I named it as such. I used this data structure so that we could easily merge into the map we're now calling allHierarchies once it's created. So, I think the name should remain.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
devchas: [This](https://blog.brendanlacanette.com/2015/08/variable-is-accessed-within-inner-class.html) is a pretty good description as to why I needed that. However, I do still think it's confusing, so I've changed this to create, update, and return the map in separate steps.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,360 @@
...
+              .fromPropertiesFile()
+              .setLoginCustomerId(loginCustomerId)
+              .build();
+    } catch (FileNotFoundException fnfe) {
+      System.err.printf(
+          "Failed to load GoogleAdsClient configuration from file. Exception: %s%n", fnfe);
+      return null;
+    } catch (IOException ioe) {
+      System.err.printf("Failed to create GoogleAdsClient. Exception: %s%n", ioe);
+      return null;
```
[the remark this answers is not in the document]
devchas: I think that's a design decision. At the outset, we decided to support partial failure, in a way, because we skip test accounts that don't produce hierarchies. There are additional reasons that a hierarchy may not be fetched (e.g. a client account with an incomplete billing setup). Going off of the initial decision to support partial failure, I've taken the following approach.

First, we get a list of all accessible customers. In the event that the GAC cannot be created, the entire example will fail in the same way most code examples fail. Once we have the list of accessible customers, we loop through each. For each one that we can retrieve an account hierarchy, we'll add it to a list of hierarchies. For each one that fails, for whatever reason, we'll add to a list of failed cases. Then, we'll print all of the IDs of the accounts that failed, followed by the hierarchies of all the IDs that were successful.

This case would be an example of a single ID failing. I don't see why this would be different than the other reasons retrieving the hierarchy might fail, so that is why I've treated it this way.

The other reasons why I chose this design are because it is parallel to how the initial GAC is created in the main method - the method just returns. In this case, we must return null because the return value is a GAC object rather than void. In addition, because we list all accessible accounts at the outset, I like the idea of covering them all in the subsequent print statements. Each customer ID will be found in either the successful cases or the failures.

All of that said, I think the main point here is that we're "supporting partial failure" for some cases, so why should this case be different and fail the entire example?

For now, I am going to leave this as as but add some more comments and also add the CID in the error message. However, let me know what you think.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -80,16 +107,110 @@ public static void main(String[] args) {
...
+    // information.
+    if (loginCustomerId != null) {
+      googleAdsClient = createGoogleAdsClient(loginCustomerId);
+    } else {
+      googleAdsClient = createGoogleAdsClient(seedCustomerId);
+    }
+
+    if (googleAdsClient == null) {
+      return null;
+    }
```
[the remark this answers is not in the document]
devchas: See my comment above.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,367 @@
...
+
+      // The rootCustomerClient will be null if the account hierarchy was unable to be retrieved
+      // (e.g. the account is a test account or a client account with an incomplete billing setup.
+      // This method returns null in these cases to add the seedCustomerId to the list of
+      // customer IDs for which the account hierarchy could not be retrieved.
+      if (rootCustomerClient == null) {
+        return null;
+      }
+
+      Map<CustomerClient, Multimap<Long, CustomerClient>> CustomerClientToHierarchy =
```
fiboknacky, line 297: Why is this variable named using a capital letter?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,367 @@
...
+        googleAdsClient.getLatestVersion().createCustomerServiceClient()) {
+      ListAccessibleCustomersResponse accessibleCustomers =
+          customerServiceClient.listAccessibleCustomers(
+              ListAccessibleCustomersRequest.newBuilder().build());
+      System.out.println(
+          "No manager customer ID is specified. The example will print the "
+              + "hierarchies of all accessible customer IDs:");
+
+      for (String customerResourceName : accessibleCustomers.getResourceNamesList()) {
+        long customer = Long.parseLong(CustomerName.parse(customerResourceName).getCustomer());
```
fiboknacky, line 323: Like the above comment, we should probably declare this as Long, not long, (and use Long.valueOf() instead) to avoid unnecessary unboxing processes.
The GoogleAdsClient mostly also uses the boxed version, so I don't see a need to use unboxed version here too.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,367 @@
...
+      googleAdsClient =
+          GoogleAdsClient.newBuilder()
+              .fromPropertiesFile()
+              .setLoginCustomerId(loginCustomerId)
+              .build();
+    } catch (FileNotFoundException fnfe) {
+      System.err.printf(
+          "Failed to load GoogleAdsClient configuration from file with loginCustomerId %d. " +
+            "Exception: %s%n", loginCustomerId, fnfe);
+      return null;
```
fiboknacky, line 180: Having a second thought, I don't think this makes sense here.
If the property file cannot be found, it should fail at the step of creating default GAC above.
Even if the IO fails somehow, we should probably fail fast and tell the example users that something is gone wrong, so they can go fix it.

In other words, what's the benefit of printing "your prop file cannot be found" a hundred times?
They still need to fix the prop file. They probably don't care that we print that "your prop file cannot be found" for all the customers that their Google account has access to.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,367 @@
...
+    if (managerId != null) {
+      seedCustomerIds.add(managerId);
+    } else {
+      seedCustomerIds = getAccessibleCustomers(googleAdsClient);
+    }
+
+    Map<CustomerClient, Multimap<Long, CustomerClient>> allHierarchies = new HashMap<>();
+    List<Long> accountsWithNoInfo = new ArrayList<>();
+    // Constructs a map of account hierarchies.
+    for (long seedCustomerId : seedCustomerIds) {
```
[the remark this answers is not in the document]
devchas: seedCustomerIds must be a list of Long because List of long is not supported. I used the primitive long in the for loop because the primitive version is all that is needed. Similarly, loginCustomerId need not be a Long, so I have changed this to a long in the params.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
devchas: Typo: fixed.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,367 @@
...
+   */
+  private GoogleAdsClient createGoogleAdsClient(long loginCustomerId) {
+    GoogleAdsClient googleAdsClient;
+    try {
+      googleAdsClient =
+          GoogleAdsClient.newBuilder()
+              .fromPropertiesFile()
+              .setLoginCustomerId(loginCustomerId)
+              .build();
+    } catch (FileNotFoundException fnfe) {
```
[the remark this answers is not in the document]
devchas: How about removing the try/catch blocks here and allowing to propagate causing the entire example to fail? That way we can avoid writing the exact same code twice in the if/else block.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,367 @@
...
+    // https://developers.google.com/google-ads/api/docs/concepts/call-structure#cid for more
+    // information.
+    if (loginCustomerId != null) {
+      googleAdsClient = createGoogleAdsClient(loginCustomerId);
+    } else {
+      googleAdsClient = createGoogleAdsClient(seedCustomerId);
+    }
+
+    // The createGoogleAdsClient method returns null if a GoogleAdsClient cannot be created. In
+    // that case, the account hierarchy for the given seedCustomerId will be able to be
```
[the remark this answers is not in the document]
devchas: This has been removed

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,367 @@
...
+    // retrieved. This method returns null in that case to add the seedCustomerId to the list of
```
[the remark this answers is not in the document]
devchas: This has been removed.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,367 @@
...
+      } else {
+        accountsWithNoInfo.add(seedCustomerId);
+      }
+    }
+
+    // Prints the IDs of any accounts that did not produce hierarchy information.
+    if (!accountsWithNoInfo.isEmpty()) {
+      System.out.println(
+          "Unable to retrieve information for the following accounts which are likely either test "
+              + "accounts with setup issues. Please check the logs for additional details.");
```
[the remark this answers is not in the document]
devchas: Fixed.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
devchas: See revised solution.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
devchas: We typically use the primitive version unless the boxed version is necessary elsewhere in the library.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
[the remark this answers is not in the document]
jradcliff: WDYT of refactoring that if/else from:
```
    if (loginCustomerId != null) {
      googleAdsClient = createGoogleAdsClient(loginCustomerId);
    } else {
      googleAdsClient = createGoogleAdsClient(seedCustomerId);
    }
```
to:
```
    googleAdsClient = createGoogleAdsClient(
        loginCustomerId != null ? loginCustomerId : seedCustomerId);
```
or:
```
    googleAdsClient = createGoogleAdsClient(
        MoreObjects.firstNonNull(loginCustomerId, seedCustomerId));
```
Then you wouldn't need this method, you still wouldn't repeat yourself, and you'd just let `createCustomerClientToHierarchy()` propagate the exception via an update to its `throws` clause.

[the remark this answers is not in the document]
jradcliff: Oops, typo...just a sec...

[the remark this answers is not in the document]
jradcliff: See updated code suggestion above.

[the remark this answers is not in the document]
devchas: Got it, fixed.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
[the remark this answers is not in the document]
fiboknacky: I meant to say that the list of Long is fine, but I'm not sure why you need primitive long here?
You still need to use seedCustomerId to create a GoogleAdsClient.Builder(), which accepts the boxed version (Long).
So, I just don't get why we want to unbox the seedCustomerId so early here, whereas the loginCustomerId is passed as the boxed version.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -114,11 +122,14 @@ public static void main(String[] args) {
    * @param managerId the root customer ID from which to begin the search.
    * @param loginCustomerId the loginCustomerId used to create the GoogleAdsClient.
    */
-  private void runExample(GoogleAdsClient googleAdsClient, Long managerId, Long loginCustomerId) {
+  private void runExample(GoogleAdsClient googleAdsClient, Long managerId, long loginCustomerId)
```
fiboknacky, line 125: Why do you change the type to long, whereas the managerId variable is still the same?
Besides, could you check with Josh or Nick?
I remember vaguely that there is a reason why we declared this using Long, not long.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -41,7 +41,13 @@
 import java.util.Map;
 import java.util.Set;
 
-/** Gets the account hierarchy of the specified manager account. */
+/**
+ * Gets the account hierarchy of the specified manager account. If you don't specify manager ID, the
```
fiboknacky, line 45: We should mention the login customer ID here too.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -114,11 +122,14 @@ public static void main(String[] args) {
    * @param managerId the root customer ID from which to begin the search.
    * @param loginCustomerId the loginCustomerId used to create the GoogleAdsClient.
    */
```
fiboknacky, line 124: Add @throw here.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
jradcliff: Our approach has been:
* If it's potentially null, then use boxed `Long`
* Otherwise, use `long`

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -42,11 +42,12 @@
 import java.util.Set;
 
 /**
- * Gets the account hierarchy of the specified manager account. If you don't specify manager ID, the
- * example will instead print the hierarchies of all accessible customer accounts for your
- * authenticated Google account. Note that if the list of accessible customers for your
- * authenticated Google account includes accounts within the same hierarchy, this example will
- * retrieve and print the overlapping portions of the hierarchy for each accessible customer.
+ * Gets the account hierarchy of the specified manager account and login customer ID. If you don't
+ * specify manager ID or login customer ID, the example will instead print the hierarchies of all
```
fiboknacky, line 46: Small question: should it be "manager ID and login customer ID"?
We want the user to provide both or none, right?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,341 @@
...
+   *
+   * @param googleAdsClient the Google Ads API client.
+   * @param managerId the root customer ID from which to begin the search.
+   * @param loginCustomerId the loginCustomerId used to create the GoogleAdsClient.
+   * @throws IOException if a Google Ads Client is not successfully created.
+   */
+  private void runExample(GoogleAdsClient googleAdsClient, Long managerId, long loginCustomerId)
+      throws IOException {
+    List<Long> seedCustomerIds = new ArrayList<>();
+    if (managerId != null) {
```
nwbirnie, line 130: nit: as a general style, I prefer leading with the positive case on if-else blocks. I just think it makes them easier to parse since it's one less boolean operator in the condition, then the not operator is naturally represented as an else block.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,341 @@
...
+              + "customer_client.id "
+              + "FROM customer_client "
+              + "WHERE customer_client.level <= 1";
+
+      // Adds the seed customer ID to the list of IDs to be processed.
+      managerAccountsToSearch.add(seedCustomerId);
+      // Performs a breadth-first search algorithm to build a mapping of managers to their
+      // child accounts.
+      Multimap<Long, CustomerClient> customerIdsToChildAccounts = ArrayListMultimap.create();
+      while (!managerAccountsToSearch.isEmpty()) {
```
nwbirnie, line 214: Here I think the logic gets a little confusing - it looks like a clearer representation here would be:

`Queue<Long> managerIdsToSearch = new ArrayBlockingQueue();`

Then pulling from the top of the queue is just `managerIdsToSearch.poll()`.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,341 @@
...
+   * Creates a map between a CustomerClient and each of its managers' mappings.
+   *
+   * @param loginCustomerId the loginCustomerId used to create the GoogleAdsClient.
+   * @param seedCustomerId the ID of the customer at the root of the tree.
+   * @return a map between a CustomerClient and each of its managers' mappings if the account
+   *     hierarchy can be retrieved. If the account hierarchy cannot be retrieved, returns null.
+   * @throws IOException if a Google Ads Client is not successfully created.
+   */
+  private Map<CustomerClient, Multimap<Long, CustomerClient>> createCustomerClientToHierarchy(
+      Long loginCustomerId, long seedCustomerId) throws IOException {
```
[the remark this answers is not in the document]
devchas: No, we have a null check in the GAC creation.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
devchas: I think the logic is pretty straight-forward here. Also you must initialize ArrayBlockingQueue with a capacity, which we don't know at the outset because we don't know how many accounts in a given hierarchy are managers.

nwbirnie: The thing I find a bit confusing here is that we're checking if the queue is empty, creating and iterator and throwing it away immediately. It looks like the Queue interface is actually what we're implementing here, not a Set.

Also sorry I just rhymed off ArrayBlockingQueue on autopilot - I think LinkedList() would be the right choice :-)

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -0,0 +1,342 @@
...
+    LinkedList<Long> managerAccountsToSearch = new LinkedList<>();
```
nwbirnie, line 185: Sorry just noticed, this should be Queue<Long>

PR #273, opened 2020-04-14
Update partial failure handling for consistency

nwbirnie:


google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadOfflineConversion.java
```diff
@@ -148,19 +148,19 @@ private void runExample(
...
-      // Prints the result.
-      ClickConversionResult result = response.getResults(0);
-      // Only prints valid results.
-      if (result.hasGclid()) {
         System.out.printf(
-            "Uploaded conversion that occurred at '%s' from Google Click ID '%s' to '%s'.%n",
-            result.getConversionDateTime().getValue(),
-            result.getGclid().getValue(),
-            result.getConversionAction().getValue());
+            "Partial error encountered: '%s'.%n", response.getPartialFailureError().getMessage());
```
jradcliff, line 152: Is this actually an if-else? If an upload has some failure and some successes, the response will have both `partialFailureError` _and_ `results`, correct?

In this example it doesn't make a difference because it's only attempting to upload one conversion, but the example should demo how users would handle multiple conversions in a request.

fiboknacky: Josh, we already have [HandlePartialFailure](https://github.com/googleads/google-ads-java/blob/master/google-ads-examples/src/main/java/com/google/ads/googleads/examples/errorhandling/HandlePartialFailure.java) that shows what you mention in detail.
That's why we (Pierrick, Nick, Fan, Devin, and I) agree that we could make this example a bit more concise. 

If that misleading is your concern, what about adding a comment at the top of the file explaining that a curious user should TAL at HandlePartialFailure?

PierrickVoulet: +1 for @fiboknacky's suggestion

jradcliff: That sounds like a reasonable compromise to me.

PierrickVoulet: I applied the change to PHP in case you want to replicate the same https://github.com/googleads/google-ads-php/pull/318/commits/5dfa4fea87327049df1ef786caa5abe80f5e6ce6

PR #274, opened 2020-04-14
Sc methods test

devchas:


google-ads/src/test/java/com/google/ads/googleads/lib/MethodsPresenceTest.java
```diff
@@ -0,0 +1,69 @@
...
+  @Test
+  public void methodsListMatches() throws ClassNotFoundException {
+    String version = "v3";
+    Class cls = Class.forName("com.google.ads.googleads." + version + ".services.GoogleAdsVersion");
+
+    // Retrieves a list of the methods (and return types) in the GoogleAdsVersion class using
+    // reflection, alphabetizes the list by method name, and concatenates the alphabetized list
+    // into a string.
+    List<Method> methods = Arrays.asList(cls.getDeclaredMethods());
+    methods.sort(Comparator.comparing((Method method) -> method.getName().toLowerCase()));
```
nwbirnie, line 51: Would it be more robust to compare based on Set equivalence, rather than relying on the input being formatted and sorted exactly as expected?

google-ads/src/test/java/com/google/ads/googleads/lib/MethodsPresenceTest.java
```diff
@@ -0,0 +1,69 @@
...
+   *
+   * <p>The avail_service_clients.txt file should be the standard against which all releases are
+   * compared. In the event the list of services changes in the API, this test will fail. Then,
+   * avail_service_clients.txt should be updated so that this test passes.
+   *
+   * <p>The version variable should be updated for each major release of the Google Ads API.
+   */
+  @Test
+  public void methodsListMatches() throws ClassNotFoundException {
+    String version = "v3";
```
nwbirnie, line 44: Could we avoid hard-coding the version? I think we should have a config file for each version, and use ApiCatalog to tell us which versions exist, and so need to be verified.

google-ads/src/test/resources/testdata/avail_service_clients.txt
```diff
@@ -0,0 +1,107 @@
+createAccountBudgetProposalServiceClient: AccountBudgetProposalServiceClient
```
nwbirnie, line 1: Would it be reasonably simple to extend this to cover method signatures as well? I'm thinking that we can use this to ensure the flattenings we're expecting are being generated correctly.

google-ads/src/test/java/com/google/ads/googleads/lib/MethodsPresenceTest.java
```diff
@@ -0,0 +1,69 @@
...
+    methods.sort(Comparator.comparing((Method method) -> method.getName().toLowerCase()));
+    String serviceListReflection = "";
+    for (Method method : methods) {
+      serviceListReflection += method.getName() + ": " + method.getReturnType().getSimpleName();
+    }
+
+    // Retrieves the list of service client methods (and return types) in the
+    // provided avail_service_clients.txt file.
+    ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+    InputStream inputStream = classLoader.getResourceAsStream("testdata/avail_service_clients.txt");
```
nwbirnie, line 60: MethodsPresenceTest.class.getResourceAsStream() is generally sufficient here - the context classloader is only required if we have multiple classloaders (e.g. in a module system). Here we only have one classloader so it should be able to see these resources.

google-ads/src/test/java/com/google/ads/googleads/lib/MethodsPresenceTest.java
```diff
@@ -0,0 +1,70 @@
...
+    InputStream inputStream =
+        MethodsPresenceTest.class.getResourceAsStream("/testdata/avail_service_clients.txt");
+    Scanner scanner = new Scanner(inputStream);
+    Set<String> serviceListResource = new HashSet<>();
+    while (scanner.hasNext()) {
+      serviceListResource.add(scanner.nextLine());
+    }
+    serviceListResource.remove("");
+
+    assertTrue(
```
nwbirnie, line 66: nit: `assertEquals(serviceListResource, serviceListReflection);`

google-ads/src/main/java/com/google/ads/googleads/lib/MethodsListGenerator.java
```diff
@@ -0,0 +1,45 @@
...
+public class MethodsListGenerator {
+  public static void main(String args[])
+      throws ClassNotFoundException, IOException, NoSuchMethodException {
+    Method latestVersionMethod = GoogleAdsAllVersions.class.getMethod("getLatestVersion");
+    Class cls = Class.forName(latestVersionMethod.getReturnType().getName());
+    List<Method> methods = Arrays.asList(cls.getDeclaredMethods());
+
+    String output = "";
+    for (Method method : methods) {
+      output += method.toString() + "\n";
```
jradcliff, line 36: I'm fairly certain that technically the `toString()` isn't needed here, but up to you if you want to keep it here for readability.

google-ads/src/main/java/com/google/ads/googleads/lib/MethodsListGenerator.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+    System.out.printf(output);
+
+    String resourcesDir = "./google-ads/src/test/resources/testdata/avail_service_clients.txt";
+    FileOutputStream classesFile = new FileOutputStream(resourcesDir);
```
jradcliff, line 41: Should use try-with-resources to ensure the output stream is closed.

**Alternative:** Use Guava I/O utils, which take care of closing files and provide a nice fluent API for common I/O tasks. See `CharSink` usage here:
https://github.com/google/guava/wiki/IOExplained

google-ads/src/test/java/com/google/ads/googleads/lib/MethodsPresenceTest.java
```diff
@@ -0,0 +1,70 @@
...
+    // Retrieves a list of the methods in the GoogleAdsVersion class using reflection and adds
+    // each to a set.
+    List<Method> methods = Arrays.asList(cls.getDeclaredMethods());
+    Set<String> serviceListReflection = new HashSet<>();
+    for (Method method : methods) {
+      serviceListReflection.add(method.toString());
+    }
+
+    // Retrieves the list of service client methods in the provided avail_service_clients.txt file.
+    InputStream inputStream =
```
jradcliff, line 57: Could collapse all this to:

```
Set<String> serviceListResource = new HashSet<String>(Resources.asCharSource(..., ...).readLines());
serviceListResource.removeAll("");
```

google-ads/src/test/java/com/google/ads/googleads/lib/MethodsPresenceTest.java
jradcliff: +1

PR #276, opened 2020-04-16
Fix minor doc typo

devchas:


google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -223,8 +223,8 @@ private void runExample(GoogleAdsClient googleAdsClient, Long managerId, Long lo
                       .setCustomerId(Long.toString(customerIdToSearchFrom))
                       .build());
 
-          // Iterates over all rows in all pages to get all customer clients under the specified
-          // customer's hierarchy.
+          // Iterates over all rows to get all customer clients under the specified customer's
```
jradcliff, line 226: The "in all pages" is accurate, in that `iterateAll()` will automatically handle fetching subsequent pages and flatten all the rows into a single iterator.

devchas: You're absolutely right. In my head, I thought I had switched this to a stream search. Just restored that language.

Issue #282, opened 2020-04-24
Add method for Ad resource name to the ResourceNames utilities


jradcliff:
**Which version of the client library are you using?**
5.0.1-SNAPSHOT

**Which version of Java are you using?**
9

**Which operating system (Linux, Windows, ...) and version?**
Linux

**Actions taken**
Using the `AdService` to update `Ad` objects.

**Expected result**
Able to use `ResourceNames.ad(customerId, adId)`.

**Actual result**
No such method exists.
 
**Anything else we should know about your project / environment**

jradcliff:
Fix merged from pull request #283 will be included in the next release of the library.

jradcliff:
Fix released in [5.1.0](https://github.com/googleads/google-ads-java/releases/tag/5.1.0).

PR #283, opened 2020-04-24
Added ad() method to ResourceNames

jradcliff:
Addresses issue #282.


jradcliff:
**NOTE:** This does not change the v1 utility because the `Ad` resource was introduced in v2.

nwbirnie:
I'm guessing we have a bug in update_gen_src_java.sh as well - most likely that we're allowing partial matches of the entity name. Is there any chance you could take a look there as well?

jradcliff:
@nwbirnie , will follow up on fixing the utility you mentioned separately. Thanks for highlighting that.

PR #286, opened 2020-04-29
Run google-java-format on all examples

nwbirnie:
This is a fully automated change produced by `google-java-format -i $(find google-ads-examples/src/main/java/com/google/ads/googleads/examples/ -type f -name *java)`


nwbirnie:
Yes, there is a hook available. I'll add this to the installation instructions on g3doc.

PR #287, opened 2020-04-29
Respect unix conventions and exit with nonzero on failure

nwbirnie:


nwbirnie:
Good point - no was missing the other 2 catches in main. There were also a couple of odd-balls outside of main which I've identified and fixed in the commit with message "Add edge cases".

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -260,6 +261,7 @@ private void runExample(GoogleAdsClient googleAdsClient, Long managerId, Long lo
           System.out.printf(
               "Unable to retrieve hierarchy for customer ID %d: %s%n",
               customerIdToSearchFrom, gae.getGoogleAdsFailure().getErrors(0).getMessage());
+          System.exit(1);
```
jradcliff, line 264: For this one, @devchas specifically set it up to return `null`. @devchas , we want to preserve that behavior and _not_ change this line, correct?

nwbirnie:
This was handled internally prior to v5 release.

PR #291, opened 2020-05-12
Adds HandleKeywordPolicyViolations example

jradcliff:
Based on PHP library example:

https://github.com/googleads/google-ads-php/blob/master/examples/ErrorHandling/HandleKeywordPolicyViolations.php

### DIFF
--- a/google-ads-examples/src/main/java/com/google/ads/googleads/examples/errorhandling/HandleKeywordPolicyViolations.java
+++ b/google-ads-examples/src/main/java/com/google/ads/googleads/examples/errorhandling/HandleKeywordPolicyViolations.java
@@ -1,0 +1,214 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.ads.googleads.examples.errorhandling;
+
+import com.beust.jcommander.Parameter;
+import com.google.ads.googleads.examples.utils.ArgumentNames;
+import com.google.ads.googleads.examples.utils.CodeSampleParams;
+import com.google.ads.googleads.lib.GoogleAdsClient;
+import com.google.ads.googleads.v3.common.KeywordInfo;
+import com.google.ads.googleads.v3.common.PolicyViolationKey;
+import com.google.ads.googleads.v3.enums.AdGroupCriterionStatusEnum.AdGroupCriterionStatus;
+import com.google.ads.googleads.v3.enums.KeywordMatchTypeEnum.KeywordMatchType;
+import com.google.ads.googleads.v3.errors.GoogleAdsError;
+import com.google.ads.googleads.v3.errors.GoogleAdsException;
+import com.google.ads.googleads.v3.errors.PolicyViolationDetails;
+import com.google.ads.googleads.v3.resources.AdGroupCriterion;
+import com.google.ads.googleads.v3.services.AdGroupCriterionOperation;
+import com.google.ads.googleads.v3.services.AdGroupCriterionServiceClient;
+import com.google.ads.googleads.v3.services.MutateAdGroupCriteriaResponse;
+import com.google.ads.googleads.v3.utils.ResourceNames;
+import com.google.protobuf.StringValue;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Demonstrates how to request an exemption for policy violations of a keyword.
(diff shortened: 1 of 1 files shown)


google-ads-examples/src/main/java/com/google/ads/googleads/examples/errorhandling/HandleKeywordPolicyViolations.java
```diff
@@ -0,0 +1,214 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Demonstrates how to request an exemption for policy violations of a keyword.
+ *
+ * <p>The example uses an exemptible policy-violating keyword by default. If you use a keyword that
+ * contains non-exemptible policy violations, they will not be sent for exemption request and you
```
devchas, line 44: s/they/it

google-ads-examples/src/main/java/com/google/ads/googleads/examples/errorhandling/HandleKeywordPolicyViolations.java
```diff
@@ -0,0 +1,214 @@
...
+ * will still fail to create a keyword.
+ *
+ * <p>If you specify a keyword that doesn't violate any policies, this example will just add the
+ * keyword as usual, similar to what the AddKeywords example does.
+ *
+ * <p>Once you've requested policy exemption for a keyword, when you send a request for adding it
```
devchas, line 50: Suggested reword: When you send a request to add a keyword after requesting a policy exemption for that keyword, the request will pass as if you were adding a non-violating keyword.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/errorhandling/HandleKeywordPolicyViolations.java
```diff
@@ -0,0 +1,214 @@
...
+      AdGroupCriterionOperation operation =
+          AdGroupCriterionOperation.newBuilder().setCreate(adGroupCriterion).build();
+
+      // Tries sending a mutate request to add the keyword.
+      List<PolicyViolationKey> exemptibleKeys = new ArrayList<>();
+      int errorCount = 0;
+      try {
+        MutateAdGroupCriteriaResponse response =
+            adGroupCriterionServiceClient.mutateAdGroupCriteria(
+                String.valueOf(customerId), Arrays.asList(operation));
```
devchas, line 148: A nit, but I think we normally use ImmuatableList.of

google-ads-examples/src/main/java/com/google/ads/googleads/examples/errorhandling/HandleKeywordPolicyViolations.java
```diff
@@ -0,0 +1,214 @@
...
+      } catch (GoogleAdsException e) {
+        exemptibleKeys = extractExemptiblePolicyViolationKeys(e);
+        errorCount = e.getGoogleAdsFailure().getErrorsCount();
+      }
+
+      // Tries sending exemption requests for creating the keyword. However, if your keyword
+      // contains many policy violations, but not all of them are exemptible, the request will not
+      // be sent.
+      if (exemptibleKeys.size() == errorCount) {
```
devchas, line 157: Shouldn't we check if the number of errors > 0?

devchas, line 157: Do we want to handle the case when `exemptibleKeys.size() != errorCount` by logging something to the user?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/errorhandling/HandleKeywordPolicyViolations.java
```diff
@@ -0,0 +1,214 @@
...
+        System.out.println(
+            "Attempting to add the keyword again by requesting exemption for its policy"
+                + " violations.");
+        // Creates a modified version of the operation with the exempt policy violation keys.
+        operation = operation.toBuilder().addAllExemptPolicyViolationKeys(exemptibleKeys).build();
+
+        // Tries sending the mutate request again.
+        MutateAdGroupCriteriaResponse response =
+            adGroupCriterionServiceClient.mutateAdGroupCriteria(
+                String.valueOf(customerId), Arrays.asList(operation));
```
devchas, line 167: Same nit as above.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/errorhandling/HandleKeywordPolicyViolations.java
jradcliff: Good point. I thought this would get a bit confusing with if/else-if/else, so instead I added an early `return` above with some comments to cover the case where the first request succeeds.

jradcliff: The call to `extractExemptiblePolicyViolationKeys()` will log all underlying errors, so I don't think we need to repeat that here.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/errorhandling/HandleKeywordPolicyViolations.java
jradcliff: Done

PR #292, opened 2020-05-13
Add example code showing how to configure the lib programatically.

nwbirnie:
Updating the documentation as part of the AIs from the UX study. This avoids having code embedded directly in markdown.

### DIFF
--- a/google-ads-examples/src/main/java/com/google/ads/googleads/examples/basicoperations/ConfigureClientLibraryWithBuilder.java
+++ b/google-ads-examples/src/main/java/com/google/ads/googleads/examples/basicoperations/ConfigureClientLibraryWithBuilder.java
@@ -1,0 +1,60 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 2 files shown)


google-ads-examples/src/main/java/com/google/ads/googleads/examples/basicoperations/ConfigureClientLibraryWithBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+            // Sets the OAuth credentials which provide Google Ads account access.
+            .setCredentials(credentials)
+            // Optional: sets the login customer ID. This is required when the Google account
+            // authenticated with the refresh token does not have direct access to
+            // OPERATING_CUSTOMER_ID and the access is via a manager account. In this case, specify
+            // the manager account ID as LOGIN_CUSTOMER_ID.
+            .setLoginCustomerId(Long.valueOf(LOGIN_CUSTOMER_ID))
+            .build();
+
+    // Now use the client configured with these credentials.
```
jradcliff, line 54: nit: "Now use" -> "Uses"

google-ads-examples/src/main/java/com/google/ads/googleads/examples/basicoperations/ConfigureClientLibraryWithBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.ads.googleads.examples.basicoperations;
+
+import com.google.ads.googleads.lib.GoogleAdsClient;
+import com.google.ads.googleads.v3.services.GoogleAdsServiceClient;
+import com.google.auth.oauth2.UserCredentials;
+
+/** Shows how to configure the client library with API access credentials with builder. */
```
jradcliff, line 21: "with API access credentials with builder" is a bit of a mouthful. Could you tweak the wording, and add a bit more info explaining that this is for the use case where you don't want to use an `ads.properties` config file and instead will provide config info via other means?

PR #296, opened 2020-05-26
Add ParallelReportDownload sample

nwbirnie:


google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+  }
+
+  /** An observer which records a simple count of the result rows received. */
+  private static class ResponseCountingObserver
+      implements ResponseObserver<SearchGoogleAdsStreamResponse> {
+
+    private final long customerId;
+    private final SettableFuture<ReportSummary> future = SettableFuture.create();
+    private AtomicLong numRecords = new AtomicLong(0);
```
jradcliff, line 176: `numResponses` would be less confusing. I misread this initially as the number of _rows_ across the responses.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -0,0 +1,249 @@
...
+    ListenableFuture<ReportSummary> asFuture() {
+      return future;
+    }
+  }
+
+  /** Summarizes the result of a reporting API call. */
+  private static class ReportSummary {
+
+    private final Long customerId;
+    private final long numRecords;
```
jradcliff, line 222: Same comment here re: `numResponses`.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    @Override
+    public void onStart(StreamController controller) {
+      // Nothing to do here.
+    }
+
+    @Override
+    public void onResponse(SearchGoogleAdsStreamResponse response) {
+      // Does something useful with the row. In this case we just count the
```
jradcliff, line 189: s/with the row/with the response/

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -0,0 +1,249 @@
...
+      // records. Note that this method may be called from multiple threads, though always in
```
jradcliff, line 190: s/records/responses

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -0,0 +1,249 @@
...
+      }
+    }
+  }
+
+  /**
+   * Runs the example. This method may block for up to MAX_WAIT_TIME_SECONDS while waiting for
+   * reports to complete.
+   *
+   * @param googleAdsClient the client library instance for API access.
+   */
```
fiboknacky, line 115: Lacks some `@param`  here.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -0,0 +1,249 @@
...
+  private void runExample(GoogleAdsClient googleAdsClient, List<Long> customerIds) {
+    // Creates a single client which can be shared by all threads.
+    // gRPC handles multiplexing parallel requests to the underlying API connection.
+    try (GoogleAdsServiceClient serviceClient =
+        googleAdsClient.getLatestVersion().createGoogleAdsServiceClient()) {
+      // IMPORTANT: You should avoid hitting the same customer ID in parallel. There are rate limit
+      // errors at the customer ID level, but not at the developer token level. Rate limits are
```
fiboknacky, line 122: >  but not at the developer token level

I'm not sure this will give impression that there is no rate limit at all for dev token level (which is not true)?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -0,0 +1,249 @@
...
+      // Does something useful with the row. In this case we just count the
+      // records. Note that this method may be called from multiple threads, though always in
+      // sequence.
+      numRecords.incrementAndGet();
+    }
+
+    @Override
+    public void onError(Throwable t) {
+      // Notify that this report failed.
+      notifyComplete(new ReportSummary(customerId, numRecords.get(), t));
```
fiboknacky, line 198: This contradicts to my intuition a bit.
When we say complete, I feel it shouldn't have any errors.
Can't we have another method named notifyError? Or can we name this method more neutral?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -0,0 +1,249 @@
...
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class ParallelReportDownload {
+
+  // Adjust these fields as required.
+  /** Defines the Google Ads Query Language (GAQL) query strings to run for each customer ID. */
+  private static final List<String> GAQL_QUERY_STRINGS =
+      ImmutableList.of(
+          "SELECT campaign.id, metrics.impressions, metrics.clicks FROM campaign",
```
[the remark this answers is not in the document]
nwbirnie: Good call - done

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -0,0 +1,249 @@
...
+    @Parameter(names = ArgumentNames.LOGIN_CUSTOMER_ID)
+    Long loginCustomerId;
+  }
+
+  public static void main(String[] args) {
+    ParallelReportDownloadParams params = new ParallelReportDownloadParams();
+    if (!params.parseArguments(args)) {
+      // Either pass the required parameters for this example on the command line, or insert them
+      // into the code here. See the parameter class definition above for descriptions.
+      params.customerIds = ImmutableList.of(Long.valueOf("INSERT CUSTOMER IDS"));
```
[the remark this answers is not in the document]
nwbirnie: Hmm, so I considered this but left it out to keep this example clearer - there's already a lot going on here and I didn't want to distract from the core message.

The `GetAccountHierarchy` example is really quite involved due to reconstructing the tree. Most of that logic isn't required here, just a simple list of customer IDs.

How about we add a separate example `ListCustomersUnderManager`? This would be short - it's just a call to select `customer_client` entries. Then we can reference that example in comments from here.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
nwbirnie: Updated wording here, PTAL.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
nwbirnie: Updated, PTAL.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -0,0 +1,249 @@
...
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class ParallelReportDownload {
```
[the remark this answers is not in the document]
nwbirnie: Will do this after resolving conversation with jradcliff re. scope of this example.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -0,0 +1,253 @@
...
+   */
+  private void runExample(GoogleAdsClient googleAdsClient, List<Long> customerIds) {
+    // Creates a single client which can be shared by all threads.
+    // gRPC handles multiplexing parallel requests to the underlying API connection.
+    try (GoogleAdsServiceClient serviceClient =
+        googleAdsClient.getLatestVersion().createGoogleAdsServiceClient()) {
+      // IMPORTANT: You should avoid hitting the same customer ID in parallel. There are rate limits
+      // at the customer ID level which are much stricter than limits at the developer token level.
+      // Hitting these limits frequently enough will significantly reduce throughput as the client
+      // library will retry with exponential back-off before eventually failing the request.
```
fiboknacky, line 129: The Java lib will retry automatically, correct?

nwbirnie: Yes, clarified.

fiboknacky:
OK. I'll wait until you resolve the discussion with Josh.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -186,26 +191,25 @@ public void onStart(StreamController controller) {
 
     @Override
     public void onResponse(SearchGoogleAdsStreamResponse response) {
-      // Does something useful with the row. In this case we just count the
-      // records. Note that this method may be called from multiple threads, though always in
-      // sequence.
-      numRecords.incrementAndGet();
+      // Does something useful with the response. In this case we just count the responses.
+      // Note: this method may be called from multiple threads, though always in sequence.
```
jradcliff, line 195: Took me a minute to understand "Though always in sequence". Consider rephrasing to clarify that you mean that responses passed here will be in the order returned by the API (I _think_ that's what you're trying to get across here).

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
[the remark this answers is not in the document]
jradcliff: Thought about this, and I see the benefits of keeping this example focused on parallel downloads instead of making it overly complicated and overly specialized for a specific use case for the source of the customer ID list.

I'm OK with keeping this as you have it, provided the class Javadoc here explicitly points the user to both the
[GetAccountHierarchy](https://github.com/googleads/google-ads-java/blob/master/google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java) 
and 
[ListAccessibleCustomers](https://github.com/googleads/google-ads-java/blob/master/google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/ListAccessibleCustomers.java)
examples as ways to retrieve a list of customer IDs for different use cases.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -0,0 +1,253 @@
...
+  /** Specifies the maximum duration that this code example may run for. */
+  private static final int MAX_WAIT_TIME_SECONDS = 60 * 60; // 1 hour.
+
+  /** Absolute wall clock time by when the job must be completed. */
+  private static final long DEADLINE_MILLIS =
+      System.currentTimeMillis() + MAX_WAIT_TIME_SECONDS * 1000;
+
+  private static class ParallelReportDownloadParams extends CodeSampleParams {
+
+    @Parameter(names = ArgumentNames.CUSTOMER_ID, required = true)
```
jradcliff, line 63: Consider adding a `description` attribute for this annotation since it's a list, and 99% of our other examples have `CUSTOMER_ID` as a single value.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -0,0 +1,253 @@
...
+      // IMPORTANT: You should avoid hitting the same customer ID in parallel. There are rate limits
+      // at the customer ID level which are much stricter than limits at the developer token level.
+      // Hitting these limits frequently enough will significantly reduce throughput as the client
+      // library will automatically retry with exponential back-off before failing the request.
+      for (String gaqlQuery : GAQL_QUERY_STRINGS) {
+        // Uses a list of futures to make sure that we wait for this report to complete on all
+        // customer IDs before proceeding. The Future data type here is just for demonstration.
+        List<ListenableFuture<ReportSummary>> futures = new ArrayList<>();
+
+        // Calls the API for each customer ID on the current report.
```
jradcliff, line 135: nit: I think the wording here's a bit confusing. Consider swapping around to something like "Uses the API to retrieve the report for each customer ID".

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    @Override
+    public void onStart(StreamController controller) {
+      // Nothing to do here.
+    }
+
+    @Override
+    public void onResponse(SearchGoogleAdsStreamResponse response) {
+      // Does something useful with the response. In this case we just count the responses.
```
jradcliff, line 194: Worth expanding a bit to something like "This implementation just counts the responses, but could also write the response to a file, pass the response on to another method for further processing, etc.."
Just want to make it clear to users that this is an intentionally simple example for readability, but they can get creative and do whatever they'd like with the results.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -0,0 +1,253 @@
...
+        // would complicate the example code and so is omitted here.
+        List<ReportSummary> results =
+            Futures.allAsList(futures)
+                .get(DEADLINE_MILLIS - System.currentTimeMillis(), TimeUnit.MILLISECONDS);
+
+        System.out.println("Report results for query: " + gaqlQuery);
+        results.forEach(System.out::println);
+      }
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
```
jradcliff, line 168: Our style guide suggests that you should either:

1. Add `InterruptedException` to `throws` or
2. Call `interrupt()` as you've done here, then wrap in a custom exception

Since the wrapping here isn't adding any info or context, I'd go for the first option instead.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
nwbirnie: sg - done

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
nwbirnie: Good point - done.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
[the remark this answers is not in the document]
nwbirnie: sgtm - added this to the class javadoc.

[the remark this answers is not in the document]
jradcliff: Looks good. Thanks for making those changes.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -242,12 +245,12 @@ boolean isSuccess() {
 
     @Override
     public String toString() {
-      return "Customer ID "
+      return "Customer ID '"
           + customerId
-          + " Number of responses: "
+          + "' Number of responses: "
```
fiboknacky: Why is there a single quote added around Customer ID and Number of responses here?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/reporting/ParallelReportDownload.java
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  /** An observer which records a simple count of the result rows received. */
+  private static class ResponseCountingObserver
+      implements ResponseObserver<SearchGoogleAdsStreamResponse> {
+
+    private final long customerId;
+    private final SettableFuture<ReportSummary> future = SettableFuture.create();
+    private final AtomicLong numResponses = new AtomicLong(0);
+    private StreamController controller;
```
fiboknacky, line 181: Not used?

nwbirnie: Good call - was testing the failure path, forgot to remove this.

PR #307, opened 2020-07-17
Changes for release GAAPI v4.

nwbirnie:


nwbirnie:
Thanks for the quick reviews folks!

PR #310, opened 2020-07-22
Gen lib with new generator, method signature changes

devchas:


nwbirnie:
The release script currently fails due to b/161541273. Workaround is available in the bug. tl;dr use JDK8 for release.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/GetAccountHierarchy.java
```diff
@@ -21,7 +21,7 @@
 import com.google.ads.googleads.v4.errors.GoogleAdsError;
```
nwbirnie, line 21: Can we change the commit message to have a bit more context about this change? I'd like to make sure that if anyone comes looking for the details, they can find our rationale for changing method signatures.

nwbirnie:
I think the first line of the commit message could be more accurate here. The original message, switching to GAPICv2 covers the full scope of this change. Then we can include that, as a side-effect, we have removed some of the overloaded method signatures which were prone to cause confusion due to inclusion of boolean flags.

PR #316, opened 2020-07-31
Add custom key/value sample for Store Sales upload

marcorosas1991:


google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadStoreSalesTransactions.java
```diff
@@ -257,7 +257,10 @@ private String createOfflineUserDataJob(
             // Continuing the example above for loyalty fraction, a value of 1.0 here indicates that
             // you are uploading all 70 of the transactions that can be identified by an email
             // address or phone number.
-            .setTransactionUploadFraction(DoubleValue.of(1.0));
+            .setTransactionUploadFraction(DoubleValue.of(1.0))
```
jradcliff, line 260: Instead of this approach, please add command line support for the key via the following steps:

1. Add a new annotated, optional field on `UploadStoreSalesTransactionsParams`. This will require a new constant in `ArgumentNames.java` as well.
1. Set the field in the `if (!params.parseArguments(args))` block above.

This is the standard approach we use in all our examples, and it makes it easier for users.

* Those who want to pass everything on the command line without editing can do so.
* Those who want to edit the example only have to edit one section.

Thanks

marcorosas1991: Thanks jradcliff. I made the changes.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadStoreSalesTransactions.java
```diff
@@ -87,6 +87,12 @@
         required = true,
         description = "The ID of a store sales conversion action")
     private Long conversionActionId;
+    
+    @Parameter(
+        names = ArgumentNames.CUSTOM_KEY,
+        required = false,
+        description = "Only required if uploading data with custom key and values")
```
nwbirnie, line 94: Can we add a bit more detail here to explain what a custom key/value pair is, when it's required and why please?

marcorosas1991: Thanks nwbirnie. I added an expanded explanation.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadStoreSalesTransactions.java
```diff
@@ -87,6 +87,14 @@
         required = true,
         description = "The ID of a store sales conversion action")
     private Long conversionActionId;
+    
+    @Parameter(
+        names = ArgumentNames.CUSTOM_KEY,
+        required = false,
+        description = "Only required after creating a custom key and custom values in the account."
+                + " Custom key and values are used to segment store sales' conversions."
```
jradcliff, line 95: nit: remove apostrophe on "sales". I don't think you need that to be possessive.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadStoreSalesTransactions.java
```diff
@@ -172,6 +183,8 @@ public static void main(String[] args)
    *     STORE_SALES_UPLOAD_THIRD_PARTY}. Otherwise, use {@code STORE_SALES_UPLOAD_FIRST_PARTY}.
    * @param externalId optional (but recommended) external ID for the offline user data job.
    * @param conversionActionId the ID of a store sales conversion action.
+   * @param customKey to segment store sales' conversions. Only required after creating a
```
jradcliff, line 186: see comment above

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadStoreSalesTransactions.java
marcorosas1991: Done.

PR #324, opened 2020-08-05
Fix handling of scenario with no provided geoTargetConstantId

devchas:
The default value of geoTargetConstantId was not being populated correctly resulting in a NullPointerError when none was provided.

### DIFF
--- a/google-ads-examples/src/main/java/com/google/ads/googleads/examples/extensions/AddGeoTarget.java
+++ b/google-ads-examples/src/main/java/com/google/ads/googleads/examples/extensions/AddGeoTarget.java
@@ -59,7 +59,7 @@ public class AddGeoTarget {
       params.feedItemId = Long.parseLong("INSERT_FEED_ID_HERE");
 
       // Optional: Specify a geoTargetConstantId.
-      params.geoTargetConstantId = GEO_TARGET_CONSTANT_ID;
+      params.geoTargetConstantId = null;
     }
 
     GoogleAdsClient googleAdsClient;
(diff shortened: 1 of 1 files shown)


google-ads-examples/src/main/java/com/google/ads/googleads/examples/extensions/AddGeoTarget.java
```diff
@@ -103,7 +103,11 @@ public static void main(String[] args) {
    * @param geoTargetConstantId the geo target constant ID to add to the extension feed item.
    */
   private void runExample(
-      GoogleAdsClient googleAdsClient, long customerId, long feedItemId, long geoTargetConstantId) {
+      GoogleAdsClient googleAdsClient, long customerId, long feedItemId, Long geoTargetConstantId) {
+    // Uses a default geoTargetConstantId value if none is provided.
+    geoTargetConstantId =
+        (geoTargetConstantId == null) ? GEO_TARGET_CONSTANT_ID : geoTargetConstantId;
```
nwbirnie, line 109: Could just set this at the field declaration, and allow the flags parser to override.

devchas: I may not be understanding you here, but I think that was the previous behavior that did not work.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/extensions/AddGeoTarget.java
```diff
@@ -59,7 +59,7 @@ public static void main(String[] args) {
       params.feedItemId = Long.parseLong("INSERT_FEED_ID_HERE");
 
       // Optional: Specify a geoTargetConstantId.
-      params.geoTargetConstantId = GEO_TARGET_CONSTANT_ID;
+      params.geoTargetConstantId = null;
```
jradcliff, line 62: Please add a `description` to the annotation on line 48 that explains that a default value will be used if not specified.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/extensions/AddGeoTarget.java
jradcliff: +1

jradcliff: I meant setting the default on line 53.

devchas: Got it. Fixed.

PR #341, opened 2020-09-21
Add sample ads.properties file

nwbirnie:
Change-Id: I3e988d37954ea20c2b4dabdf99dfe4c43d506bde

### DIFF
--- a/ads.properties.sample
+++ b/ads.properties.sample
@@ -2,6 +2,9 @@
 #
 # Defines configuration for Google Ads API client library for Java.
 #
+# See our developer documentation for more details:
+# https://developers.google.com/google-ads/api/docs/client-libs/java/quick-start
+#
 # - By default this is searched for in ~/ads.properties.
 #   e.g. GoogleAdsClient.newBuilder().fromPropertiesFile() 
 #
@@ -28,9 +31,8 @@
 #
 # Specifies the client secret that is used for Google's OAuth 2 API. 
 #
-# - Must match the OAuth clientId. 
-# - Created and managed via http://console.developers.google.com
-# - Provides an additional layer of security on top of refresh token.
+# - Must be the secret associated with the OAuth clientId.
+# - Created and managed via http://console.developers.google.com.
 #api.googleads.clientSecret=
 
 # OAuth refresh token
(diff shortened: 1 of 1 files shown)


devchas (commented):
I would suggest uncommenting the required lines and adding INSERT_ITEM. For example

api.googleads.clientId=INSERT_CLIENT_ID

nwbirnie:
> I would suggest uncommenting the required lines and adding INSERT_ITEM. For example
> 
> api.googleads.clientId=INSERT_CLIENT_ID

There's technically no required fields though... anything can be configured in code or config file or both. 

I also think that it's more helpful not to put placeholders in here. I don't think anyone would have questions about where to put the details. It just means that people need to remove these before they can use the template.

devchas (approved):

ads.properties.sample
```diff
@@ -0,0 +1,49 @@
+# OAuth refresh token
+#
+# Credential which authorizes access to a Google account. 
+#
+# - See guide for instructions on how to obtain a refresh token.
+#   https://developers.google.com/google-ads/api/docs/oauth/client-library
+# - Can access many Google Ads accounts. 
+# - Operating account is specified by the customerId field in each request. 
+#api.googleads.refreshToken=
+
```
jradcliff, line 17: We should add:

```
#api.googleads.loginCustomerId=
```

I would think a significant portion of users use credentials from a single manager account and just want to set and forget this value.

ads.properties.sample
```diff
@@ -0,0 +1,49 @@
...
+#   accounts. 
+#api.googleads.developerToken=
+
+# Endpoint
+#
+# Specifies the endpoint to use for connecting to the ads API. 
+#
+# - Can be used to connect to a staging endpoint. 
+# - Most external clients can ignore this field.
+#api.googleads.endpoint=
```
jradcliff, line 80: Have you seen this prove useful for users? If not, I'd prefer we leave it out to avoid confusion.

ads.properties.sample
```diff
@@ -0,0 +1,49 @@
+# OAuth refresh token
```
jradcliff: Add an intro comment that explains what this file is and points to the [Getting started guide for Java](https://developers.google.com/google-ads/api/docs/client-libs/java/getting-started).

Also, friendly reminder to update that guide to point to this sample file if you think that'd be useful.

jradcliff (changes requested):

ads.properties.sample
nwbirnie: We already have a section for LCID, PTAL.

ads.properties.sample
nwbirnie: Added to the intro and mailed cl/334568432

ads.properties.sample
jradcliff: Sorry, don't know how I missed that...

jradcliff (approved):

Issue #344, opened 2020-09-23
FieldMasks Utility Does Not Properly Detect Campaign BiddingStrategy of MANUAL_CPC and EnhancedCpcEnabled = false


PeterLavetsky:
Hi All,

This bug may partially be related to https://github.com/googleads/google-ads-java/issues/272

Setting the ManualCpc object on a Campaign with enhancedCpcEnabled = false, the FieldMask utility does not add the manual_cpc.enhanced_cpc_enabled path to the update_mask, and the change does not take effect on the mutate.

When enhancedCpcEnabled = true the FieldMask utility does add manual_cpc.enhanced_cpc_enabled to the update_mask and the intended change does take effect.

```
final ManualCpc manualCpc = ManualCpc.newBuilder()
                .setEnhancedCpcEnabled( false )
                .build();

googleCampaign.setManualCpc( manualCpc );

final FieldMask.Builder fieldMaskBuilder = FieldMasks.allSetFieldsOf( googleCampaign ).toBuilder();

operationBuilder
                    .setUpdate( googleCampaign )
                    .setUpdateMask( fieldMaskBuilder.build() );


```

Our current workaround:

```
if( fieldMaskBuilder.getPathsList().contains( "bidding_strategy_type" ) ) {
                    if( googleCampaign.hasManualCpc() && !fieldMaskBuilder.getPathsList().contains( "manual_cpc.enhanced_cpc_enabled" ) ) {
                        fieldMaskBuilder.addPaths( "manual_cpc.enhanced_cpc_enabled" );
                    }
                }
```

Does Not Work Request-id : zbMkzQ8qndiYjz6cVa2SHg
Works Request-id : A6epI0FMrpTS2kOu6a2wKw

The full requests are shown below, but here are the diffs of two identical requests of the same campaign with the only difference being the enhanced_cpc_enabled flag is either true or false:

```
peterlavetsky@Harpua ~ $ diff works.txt does-not-work.txt 
0a1
> 
31c32
<       enhanced_cpc_enabled: true
---
>       enhanced_cpc_enabled: false
52d52
<     paths: "manual_cpc.enhanced_cpc_enabled"
peterlavetsky@Harpua ~ $ 
```

Relevant requests with IDs mostly obfuscated below. Can be provided if needed tho I believe the above request-ids should be enough.

Does not work:

```

Request
-------
MethodName: google.ads.googleads.v5.services.CampaignService/MutateCampaigns
Endpoint: googleads.googleapis.com:443
[...]

PR #345, opened 2020-09-24
Fix field masks for presence with default value #344

nwbirnie:
Change-Id: Ic4e0f17586647d482f444f9360d6f1fc09018aa7

### DIFF
--- a/google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
+++ b/google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
@@ -17,7 +17,6 @@ package com.google.ads.googleads.lib.utils;
 import com.google.common.base.Preconditions;
 import com.google.protobuf.Descriptors.Descriptor;
 import com.google.protobuf.Descriptors.FieldDescriptor;
-import com.google.protobuf.Descriptors.FieldDescriptor.Type;
 import com.google.protobuf.FieldMask;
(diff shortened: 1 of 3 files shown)


devchas (approved):

google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
```diff
@@ -76,12 +75,14 @@ private static void compare(
           mask.addPaths(fieldName);
         }
       } else {
+        boolean hasValueChanged =
+            original.hasField(field) != modified.hasField(field)
```
jradcliff, line 79: Do the calls to `original.getField(field)` and `modified.getField(field)` above change the presence bit? I see those return the default value if the field is not set, but I don't know if they have any side effects.

nwbirnie: Good question, but no since protobuf messages in Java are immutable.

jradcliff: Thanks, this change LGTM. Going forward we may need to revisit since the protobuf team recently was discouraging the use of `hasField()` checks, but I only expect that to become a concern when and if we remove the `optional` keyword on mutable fields.

jradcliff (approved):

Issue #332, opened 2020-09-05
Don't Lose Exception Details When Wrapped Into com.google.api.gax.rpc.InvalidArgumentException


PeterLavetsky:
With DEBUG logging on for com.google.ads.googleads.lib.request.detail I can see the following request and response:

```
Request
-------
MethodName: google.ads.googleads.v4.services.GoogleAdsService/SearchStream
Endpoint: googleads.googleapis.com:443
Headers: {developer-token=REDACTED, login-customer-id=ABC, x-goog-api-client=gl-java/1.8.0_161 gapic/ gax/1.57.0 grpc/1.30.0}
Body: customer_id: "XYZ"
query: "select customer.id, campaign_budget.id, metrics.cost_micros, metrics.clicks, metrics.conversions, metrics.impressions, campaign_budget.period, campaign_budget.reference_count, campaign_budget.amount_micros, campaign_budget.name from campaign_budget where segments.date   DURING   THIS_MONTH "
```

```
Response
--------
Headers: Metadata(content-type=application/grpc,request-id=wKbcK0j7Cw3Kkp-CdTm8zQ,date=Sat, 05 Sep 2020 14:47:42 GMT,alt-svc=h3-29=":443"; ma=2592000,h3-27=":443"; ma=2592000,h3-T051=":443"; ma=2592000,h3-T050=":443"; ma=2592000,h3-Q050=":443"; ma=2592000,h3-Q046=":443"; ma=2592000,h3-Q043=":443"; ma=2592000,quic=":443"; ma=2592000; v="46,43")
Body: null
Failure message: errors {
  error_code {
    query_error: REQUESTED_METRICS_FOR_MANAGER
  }
  message: "Metrics cannot be requested for a manager account. To retrieve metrics, issue separate requests against each client account under the manager account."
}
```

However this gets wrapped into a thrown exception of com.google.api.gax.rpc.InvalidArgumentException and as far as I can tell there's no way to unpack this exception to get any valuable details ie.:

```
detailMessage : io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Request contains an invalid argument.
```

The true error should be exposed in the thrown Exception so adequate handling may be done.

Thanks
Pete

Java Client Library : com.google.api-ads:google-ads:8.0.0

nwbirnie:
[...]

PR #349, opened 2020-10-06
Transform exceptions on server streaming calls.

nwbirnie:
Currently we do not generate GoogleAdsException instances for server streaming callables (see #332). This means that users cannot retrieve the GoogleAdsFailure objects when streaming calls fail.

This change adds an ExceptionTransformingServerStreamingCallable class which handles the exception transformation. It also adds unit tests for both the new and the existing ExceptionTransformingUnaryCallable classes.

The remaining callable types will be handled in a follow up CL.

Change-Id: Ia8339f1ee482ab7f02085f5c47f90ab1cc0aee5b


google-ads/src/main/java/com/google/ads/googleads/lib/callables/ExceptionTransformation.java
```diff
@@ -0,0 +1,7 @@
+package com.google.ads.googleads.lib.callables;
+
+import com.google.api.gax.rpc.ApiException;
+
+public interface ExceptionTransformation {
+  Throwable transform(ApiException throwable);
```
jradcliff, line 6: `ApiException` is a `RuntimeException`, but here we are "escalating" it to a `Throwable`, which means any exception returned by this transformation needs to be caught and handled. Is that necessary?

google-ads/src/main/java/com/google/ads/googleads/lib/callables/ExceptionTransformation.java
```diff
@@ -0,0 +1,7 @@
+package com.google.ads.googleads.lib.callables;
+
+import com.google.api.gax.rpc.ApiException;
+
+public interface ExceptionTransformation {
```
jradcliff, line 5: Do you anticipate adding more to this interface? If not, you could simply use `java.util.function.Function<ApiException, Throwable>` or, if you'd prefer to have a named interface, just have this `extend java.util.function.Function<ApiException, Throwable>`.

This will also make these objects play nice with other functional programming utilities, etc.

google-ads/src/main/java/com/google/ads/googleads/lib/callables/ExceptionTransformingServerStreamingCallable.java
```diff
@@ -0,0 +1,66 @@
...
+    this.exceptionTransformation = exceptionTransformation;
+  }
+
+  @Override
+  public void call(
+      RequestT request, ResponseObserver<ResponseT> responseObserver, ApiCallContext context) {
+    innerCallable.call(request, new ExceptionTransformingStreamObserver(responseObserver), context);
+  }
+
+  private class ExceptionTransformingStreamObserver implements ResponseObserver<ResponseT> {
```
jradcliff, line 35: A bit of javadoc here would be helpful.

google-ads/src/test/java/com/google/ads/googleads/lib/callables/ExceptionTransformingServerStreamingCallableTest.java
```diff
@@ -0,0 +1,103 @@
...
+      }
+      GoogleAdsException exception = (GoogleAdsException) argument;
+      return Objects.equals(googleAdsFailure, exception.getGoogleAdsFailure());
+    }
+  }
+
+  private static class MockServerStreamingCallable<RequestT, ResponseT>
+      extends ServerStreamingCallable<RequestT, ResponseT> {
+
+    private volatile ResponseObserver<ResponseT> responseObserver;
```
jradcliff, line 91: nit: Is there any multithreading happening that makes this `volatile` necessary?

google-ads/src/test/java/com/google/ads/googleads/lib/callables/ExceptionTransformingServerStreamingCallableTest.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Objects;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentMatcher;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ExceptionTransformingServerStreamingCallableTest {
```
jradcliff, line 29: Please add some comments to help readers

google-ads/src/test/java/com/google/ads/googleads/lib/callables/ExceptionTransformingServerStreamingCallableTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    rawApiException =
+        new ApiException(
+            new StatusException(Status.INVALID_ARGUMENT, trailers),
+            GrpcStatusCode.of(Code.INVALID_ARGUMENT),
+            false);
+  }
+
+  @Test
+  public void transformsExceptions() {
```
jradcliff, line 61: We're testing the nice path here, but are there cases where the API will place an exception that is not transform-able in the trailing metadata key? If so, we should test for that here as well.

google-ads/src/test/java/com/google/ads/googleads/lib/callables/ExceptionTransformingServerStreamingCallableTest.java
```diff
@@ -0,0 +1,103 @@
+package com.google.ads.googleads.lib.callables;
```
jradcliff, line 1: Please add copyright comments here and on other new files

google-ads/src/test/java/com/google/ads/googleads/lib/callables/ExceptionTransformingUnaryCallableTest.java
```diff
@@ -0,0 +1,81 @@
...
+import io.grpc.StatusException;
+import java.util.concurrent.ExecutionException;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ExceptionTransformingUnaryCallableTest {
```
jradcliff, line 32: Comments on [ExceptionTransformingServerStreamingCallableTest.java](google-ads/src/test/java/com/google/ads/googleads/lib/callables/ExceptionTransformingServerStreamingCallableTest.java) apply here as well.

jradcliff (changes requested):

google-ads/src/main/java/com/google/ads/googleads/lib/callables/ExceptionTransformation.java
nwbirnie: Good point - the GAX interfaces here all deal with throwables. However, the GAPIC interfaces actually don't expose checked exceptions, everything gets rethrown as an ApiException. I've switched to deal with Throwables instead, PTAL.

google-ads/src/main/java/com/google/ads/googleads/lib/callables/ExceptionTransformation.java
nwbirnie: I think this interface is important enough to warrant its own definition. We're going to implement this 5 times for each type of RPC the library supports. Passing this around in a Function<> seems like primitive obsession.

google-ads/src/test/java/com/google/ads/googleads/lib/callables/ExceptionTransformingServerStreamingCallableTest.java
nwbirnie: Nope, removed.

google-ads/src/test/java/com/google/ads/googleads/lib/callables/ExceptionTransformingServerStreamingCallableTest.java
nwbirnie: Added some test cases.

nwbirnie:
Sorry for the force push, forgot that messes up git's review history.

google-ads/src/main/java/com/google/ads/googleads/lib/callables/ExceptionTransformation.java
jradcliff: Could you still have it extend `Function<ApiException, Throwable>`? That'll make it usable with other FP features of Java later, and I don't see a strong reason against it.

google-ads/src/test/java/com/google/ads/googleads/lib/GoogleAdsClientTest.java
```diff
@@ -450,19 +419,43 @@ public void exceptionTransformedToGoogleAdsException() {
...
+    trailers.put(trailerKey, failure.build().toByteArray());
+    StatusException rootCause = new StatusException(Status.UNKNOWN, trailers);
+    mockService.addException(new ApiException(rootCause, GrpcStatusCode.of(Code.UNKNOWN), false));
+    try (GoogleAdsServiceClient googleAdsServiceClient =
+        client.getLatestVersion().createGoogleAdsServiceClient()) {
+      googleAdsServiceClient
+          .searchStreamCallable()
+          .call(SearchGoogleAdsStreamRequest.getDefaultInstance());
+    } catch (GoogleAdsException ex) {
+      // Expected
```
jradcliff, line 446: since this is expected, then please add a `fail()` call at the end of the `try` block

google-ads/src/main/java/com/google/ads/googleads/lib/callables/ExceptionTransformingServerStreamingCallable.java
jradcliff: Did anything change on this one?

google-ads/src/main/java/com/google/ads/googleads/lib/callables/ExceptionTransformingUnaryCallable.java
```diff
@@ -80,7 +79,9 @@ public void onFailure(Throwable throwable) {
       if (throwable instanceof CancellationException && cancelled) {
         // this just circled around, so ignore.
       } else if (throwable instanceof ApiException) {
-        setException(transformation.transform((ApiException) throwable));
+        setException(transformation.transform(throwable));
```
jradcliff, line 82: (continuing the earlier comment here after the force push)

With this check, do you still need `transformation.transform()` to use `Throwable` in its return type and args?

google-ads/src/main/java/com/google/ads/googleads/lib/callables/ExceptionTransformation.java
nwbirnie: So this is already compatible with `Function<ApiException, Throwable>`. For example the following code already compiles:

```
    Function<Throwable, Throwable> foo = exceptionTransformation::transform;
```

However, I don't think that this should implement that interface for a couple reasons (1) it's not actually used/needed anywhere, (2) it constrains the future development of this interface to support the functional contract. Happy to change this if there's something I'm missing here?

google-ads/src/main/java/com/google/ads/googleads/lib/callables/ExceptionTransformingServerStreamingCallable.java
nwbirnie: Sorry missed this. Done now.

google-ads/src/test/java/com/google/ads/googleads/lib/GoogleAdsClientTest.java
nwbirnie: Good call, done. Also needed to add a loop to attempt to process the results.

google-ads/src/main/java/com/google/ads/googleads/lib/callables/ExceptionTransformingUnaryCallable.java
nwbirnie: I think this is the right type for this interface: we're given Throwable everywhere by GAX, so from a consistency point of view it seems correct to process throwable here. I agree that in general we want to avoid dealing with throwable.

google-ads/src/main/java/com/google/ads/googleads/lib/callables/ExceptionTransformation.java
jradcliff: Sounds good. I didn't realize that code snippet you provided will work, which was my main concern.

jradcliff (approved):

devchas (approved):

PR #352, opened 2020-10-12
Update installed application -> desktop application

nwbirnie:


google-ads-examples/src/main/java/com/google/ads/googleads/examples/authentication/AuthenticateInDesktopApplication.java
```diff
@@ -30,12 +30,11 @@
...
 
 /**
- * Creates an OAuth2 refresh token for the Google Ads API using the Native/Installed application
- * flow.
+ * Creates an OAuth2 refresh token for the Google Ads API using the Desktop application flow.
  *
  * <p>This example is meant to be run from the command line and requires user input.
  */
-public class AuthenticateInStandaloneApplication {
+public class AuthenticateInDesktopApplication {
```
jradcliff, line 37: WDYT of keeping this as-is? We chose the term "standalone" to try to avoid renaming the example due to minor terminology changes like this.

nwbirnie: Hmm, so it's definitely preferable to match the wording used in the developer documentation, particularly for new users who are not familiar with the terminology. I would guess that this changes every couple of years. If you compare this to the ~ 30 minutes of our time to update when the terminology changes I think it's a reasonable trade-off. WDYT?

jradcliff: Yeah, after thinking about it overnight I agree.

jradcliff (approved):

AnashOommen (approved):

Issue #353, opened 2020-10-12
java.util.concurrent.TimeoutException: Waited seconds for com.google.api.gax.retrying.CallbackChainRetryingFuture@7ec02838


din-heg-sie:
I am running the AddCustomerMatchUserList class from the com.google.ads.googleads.examples.remarketing
I am  using the  OfflineUserDataJobs  and OfflineUserDataJobService to upload data for user lists asynchronously. 

After I submit the job I am getting a timeout error while I wait for the results of my job.

**Which version of the client library are you using?**
Google Ads API v5

**Which version of Java are you using?**
jdk 1.8

**Which operating system (Linux, Windows, ...) and version?**
Windows

**Actions taken**
In creased the time out to 15 minutes. I still get the error message.

I am using the offline to submit  OfflineUserDataJobService operations. After submitting the offline job when I invoke runFuture.getPollingFuture().get(MAX_TOTAL_POLL_INTERVAL_SECONDS,  TimeUnit.SECONDS);
I get this error message:
java.util.concurrent.TimeoutException: Waited 60 seconds (plus 69870 nanoseconds delay) for com.google.api.gax.retrying.CallbackChainRetryingFuture@7ec02838[status=PENDING] at shaded.com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:495) at com.google.ads.googleads.examples.remarketing.AddCustomerMatchUserList.addUsersToCustomerMatchUserList(AddCustomerMatchUserList.java:320) at com.google.ads.googleads.examples.remarketing.AddCustomerMatchUserList.runExample(AddCustomerMatchUserList.java:198) at com.google.ads.googleads.examples.remarketing.AddCustomerMatchUserList.run(AddCustomerMatchUserList.java:161)

**Expected result**
The getPollingFuture() method should detect that the offline job submitted completed and exit smoothly without any timeout exception.

**Actual result**
Exception in thread "main" java.util.concurrent.TimeoutException: Waited 60 seconds (plus 285000 nanoseconds delay) for com.google.api.gax.retrying.CallbackChainRetryingFuture@6f7a20da[status=PENDING]
	at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:495)
	at com.google.ads.googleads.examples.remarketing.AddCustomerMatchUserListOffLine.addUsersToCustomerMatchUserList(AddCustomerMatchUserListOffLine.java:270)
[...]

PR #357, opened 2020-10-20
Update messaging in AddCustomerMatchUserList example

devchas:
In response to #353

### DIFF
--- a/google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddCustomerMatchUserList.java
+++ b/google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddCustomerMatchUserList.java
@@ -62,7 +62,7 @@ import java.util.List;
  * <p><em>Notes:</em>
  *
  * <ul>
- *   <li>This feature is only available to whitelisted accounts. See
+ *   <li>This feature is only available to allowlisted accounts. See
  *       https://support.google.com/adspolicy/answer/6299717 for more details.
  *   <li>It may take up to several hours for the list to be populated with members.
  *   <li>Email addresses must be associated with a Google account.
--- a/google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddCustomerMatchUserList.java
+++ b/google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddCustomerMatchUserList.java
@@ -256,6 +256,8 @@ public class AddCustomerMatchUserList {
       System.out.println("Waiting until operation completes.");
 
       // The polling future implements a default back-off policy for retrying.
+      // Note: Retrieving the results of the asynchronous request is for demonstrative purposes only
+      // as it may take several hours for the user list to be populated with the users.
       runFuture.getPollingFuture().get(MAX_TOTAL_POLL_INTERVAL_SECONDS, TimeUnit.SECONDS);
       System.out.printf(
           "Offline user data job with resource name '%s' has finished.%n",
--- a/google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddCustomerMatchUserList.java
+++ b/google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddCustomerMatchUserList.java
@@ -257,7 +257,7 @@ public class AddCustomerMatchUserList {
 
       // The polling future implements a default back-off policy for retrying.
       // Note: Retrieving the results of the asynchronous request is for demonstrative purposes only
-      // as it may take several hours for the user list to be populated with the users.
+      // as the job may not complete for several hours.
       runFuture.getPollingFuture().get(MAX_TOTAL_POLL_INTERVAL_SECONDS, TimeUnit.SECONDS);
       System.out.printf(
           "Offline user data job with resource name '%s' has finished.%n",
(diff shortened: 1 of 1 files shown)


google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddCustomerMatchUserList.java
```diff
@@ -256,6 +256,8 @@ private void addUsersToCustomerMatchUserList(
       System.out.println("Waiting until operation completes.");
 
       // The polling future implements a default back-off policy for retrying.
+      // Note: Retrieving the results of the asynchronous request is for demonstrative purposes only
+      // as it may take several hours for the user list to be populated with the users.
```
jradcliff, line 260: nit: replace

"as it may take several hours for the user list to be populated with the users" 

with 

"as the job may not complete for several hours"

This converts to active instead of passive voice, and avoids the common misunderstanding that all users are guaranteed to match and wind up in the list.

jradcliff (changes requested):

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddCustomerMatchUserList.java
```diff
@@ -256,6 +256,8 @@ private void addUsersToCustomerMatchUserList(
       System.out.println("Waiting until operation completes.");
 
       // The polling future implements a default back-off policy for retrying.
+      // Note: Retrieving the results of the asynchronous request is for demonstrative purposes only
```
nwbirnie, line 259: Can we give guidance for how to restart polling asynchronously? It seems like this could be a source of confusion...

jradcliff: I experimented with getting the LRO handle again when I worked on the Stores Sales upload example, and found that process confusing as well. This is why that example just prints out a GAQL query the user can use to come back and check on the results of the LRO later:

https://github.com/googleads/google-ads-java/blob/master/google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadStoreSalesTransactions.java#L494

WDYT of going with that approach here?

nwbirnie: SGTM

devchas: SGTM as well. I've also added an option to print the results if the status is SUCCESS.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddCustomerMatchUserList.java
```diff
@@ -312,6 +291,85 @@ private void addUsersToCustomerMatchUserList(
...
+          "Offline user data job ID %d with type '%s' has status: %s%n",
+          offlineUserDataJob.getId().getValue(),
+          offlineUserDataJob.getType(),
+          offlineUserDataJob.getStatus());
+      OfflineUserDataJobStatus jobStatus = offlineUserDataJob.getStatus();
+      if (OfflineUserDataJobStatus.SUCCESS == jobStatus) {
+        // Prints information about the user list.
+        printCustomerMatchUserListInfo(googleAdsClient, customerId, userListResourceName);
+      } else if (OfflineUserDataJobStatus.FAILED == jobStatus) {
+        System.out.printf("  Failure reason: %s%n", offlineUserDataJob.getFailureReason());
```
nwbirnie, line 361: Is there a good way to get more details about what failed? E.g. can we show how to get the failed operations?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddCustomerMatchUserList.java
```diff
@@ -56,9 +55,6 @@
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 
 /**
```
jradcliff, line 59: Please update the first bullet point to say "allowlisted"

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddCustomerMatchUserList.java
devchas: If the adding of operations to the job fails, there is a partial failure above that will give more specific information. This failure represents an error in the processing of that job, and I believe that this is as detailed a failure as we can provide.

jradcliff: @devchas is correct. By design, the job does not report whether each member operation succeeded or failed.

jradcliff (approved):

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/AddCustomerMatchUserList.java
nwbirnie: Ah yes, makes sense.

nwbirnie (approved):

PR #376, opened 2020-12-08
Add PII scrubbing for Feed

nwbirnie:
Change-Id: I74411b13798a96f251f0e07ac0da1af84796bce6

### DIFF
--- a/google-ads/src/main/java/com/google/ads/googleads/lib/logging/scrub/AbstractScrubber.java
+++ b/google-ads/src/main/java/com/google/ads/googleads/lib/logging/scrub/AbstractScrubber.java
@@ -1,0 +1,73 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.ads.googleads.lib.logging.scrub;
+
+import com.google.ads.googleads.lib.catalog.ApiCatalog;
+import com.google.ads.googleads.lib.catalog.Version;
+import com.google.ads.googleads.lib.utils.messageproxy.MessageEditor;
+import com.google.protobuf.Message;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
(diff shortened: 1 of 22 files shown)


google-ads/src/main/java/com/google/ads/googleads/lib/logging/scrub/AbstractScrubber.java
```diff
@@ -0,0 +1,73 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Abstracts the functionality of scrubbing a message. Handles pinning the scrub to a minimum
+ * version.
+ */
+public abstract class AbstractScrubber implements MessageEditor<Message> {
+
+  private static final ApiCatalog CATALOG = ApiCatalog.getDefault();
```
jradcliff, line 31: I'd prefer CATALOG was just a plain `private final` field, and that we pass it in the constructor. Then this class isn't making any assumptions about the life cycle or singleton-ness of what `getDefault()` returns.

This could also make testing easier if you wanted to, for example, pass a custom catalog for a test.

nwbirnie: This isn't being used though.... so we'd be adding a constructor overload that's not used anywhere... either that or we duplicate the getDefault call into all the subclasses, but duplicating this everywhere doesn't really seem to make sense.

Either way, I've made this change in the interest of brevity.

devchas (approved):

google-ads/src/main/java/com/google/ads/googleads/lib/logging/scrub/AbstractScrubber.java
jradcliff: What I meant was that the caller of the constructor should pass in the Catalog instead of the base class hard-coding it, so each subclass would have a public constructor that takes an `ApiCatalog`. The scrubber shouldn't care where the catalog comes from.

With that approach, the `LogScrubber` would pass the catalog down to each constructor it creates. I'm fine with what you have here now since the number of `ApiCatalog.getDefault()` occurrences is the same either way, but I think we do need to be mindful of the number of times we have objects fetch dependencies instead of getting passed dependencies. Doing the former makes it more difficult to refactor things later, particularly with static state like the catalog.

jradcliff (approved):

google-ads/src/main/java/com/google/ads/googleads/lib/logging/scrub/AbstractScrubber.java
nwbirnie: Sure I understand... do you think that we should have a goal of dependency inversion everywhere? It seems to me that we should be pragmatic about dependency inversion, since the code becomes harder to read/maintain/debug.

jradcliff: In my experience, dependency injection often makes the code easier to read/maintain/debug, but I realize opinions vary on that. Let's have that discussion separately.

PR #385, opened 2020-12-09
Use readable datetime instead of timestamp in resource names of code examples (Fix #42)

### LINKED ISSUE #42: Generate more useful object names in examples
Examples such as [AddCampaigns.java](https://github.com/googleads/google-ads-java/blob/master/google-ads-examples/src/main/java/com/google/ads/googleads/examples/basicoperations/AddCampaigns.java#L108) use `System.currentTimeMillis()` to generate random names. This is _somewhat_ useful, but is there an approach we could take to make these names human-readable? For example, if we used `yyyyMMdd_HH:mm:ss` in **UTC**, then users could easily identify _when_ they ran each example or identify the most recently created objects.

PierrickVoulet:
This uses the same granularity as before (milliseconds) to avoid any regression related to concurrent name conflicts.

Two formats are introduced:

1. Default: a resource name typically looks like the following `Interplanetary Cruise #2020-12-09T14:27:32.388-05:00`.
1. Short: for the resources that requires a short name, a resource name typically looks like the following `Feed name  1209142732388`.

### DIFF
--- a/google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddAdCustomizer.java
+++ b/google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddAdCustomizer.java
@@ -14,9 +14,10 @@
 
 package com.google.ads.googleads.examples.advancedoperations;
 
+import static com.google.ads.googleads.examples.utils.CodeSampleHelper.getPrintableDatetime;
+
 import com.beust.jcommander.Parameter;
 import com.google.ads.googleads.examples.utils.ArgumentNames;
-import com.google.ads.googleads.examples.utils.CodeSampleHelper;
 import com.google.ads.googleads.examples.utils.CodeSampleParams;
 import com.google.ads.googleads.lib.GoogleAdsClient;
 import com.google.ads.googleads.v6.common.ExpandedTextAdInfo;
@@ -142,7 +143,7 @@ public class AddAdCustomizer {
           "Please pass exactly two ad group IDs in the adGroupId parameter.");
     }
 
-    String feedName = "Ad Customizer example feed " + CodeSampleHelper.getPrintableDatetime();
+    String feedName = "Ad Customizer example feed " + getPrintableDatetime();
 
     // Create a feed to be used as the ad customizer.
     String adCustomizerFeedResourceName =
--- a/google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddAdCustomizer.java
+++ b/google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddAdCustomizer.java
@@ -16,6 +16,7 @@ package com.google.ads.googleads.examples.advancedoperations;
 
 import com.beust.jcommander.Parameter;
 import com.google.ads.googleads.examples.utils.ArgumentNames;
+import com.google.ads.googleads.examples.utils.CodeSampleHelper;
 import com.google.ads.googleads.examples.utils.CodeSampleParams;
 import com.google.ads.googleads.lib.GoogleAdsClient;
 import com.google.ads.googleads.v6.common.ExpandedTextAdInfo;
@@ -141,7 +142,7 @@ public class AddAdCustomizer {
           "Please pass exactly two ad group IDs in the adGroupId parameter.");
     }
 
-    String feedName = "Ad Customizer example feed " + System.currentTimeMillis();
+    String feedName = "Ad Customizer example feed " + CodeSampleHelper.getPrintableDatetime();
 
     // Create a feed to be used as the ad customizer.
     String adCustomizerFeedResourceName =
--- a/google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddAdCustomizer.java
+++ b/google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddAdCustomizer.java
@@ -14,7 +14,7 @@
 
 package com.google.ads.googleads.examples.advancedoperations;
 
-import static com.google.ads.googleads.examples.utils.CodeSampleHelper.getPrintableDatetime;
+import static com.google.ads.googleads.examples.utils.CodeSampleHelper.getShortPrintableDatetime;
 
 import com.beust.jcommander.Parameter;
 import com.google.ads.googleads.examples.utils.ArgumentNames;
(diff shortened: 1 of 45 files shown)


google-ads-examples/src/main/java/com/google/ads/googleads/examples/utils/CodeSampleHelper.java
```diff
@@ -0,0 +1,31 @@
...
+/** A helper class for all code examples. */
+public abstract class CodeSampleHelper {
+
+  /**
+   * Generates a printable string for the current date and time in UTC time zone.
+   *
+   * @return the result string.
+   */
+  public static String getPrintableDatetime() {
+    return new DateTime(DateTimeZone.UTC).toString("yyyyMMdd_HH:mm:ss.SSS") + "_UTC";
```
nwbirnie, line 29: WDYT about using the timezone of the machine so that the names are familiar to the user who ran the example? I'm thinking it could be confusing to look in the UI and see timestamps from another timezone.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/utils/CodeSampleHelper.java
```diff
@@ -0,0 +1,31 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.ads.googleads.examples.utils;
+
+import org.joda.time.DateTime;
```
nwbirnie, line 17: No need to use joda after Java 8. It looks like you want [`ZonedDateTime`](https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html).

PierrickVoulet: Not sure that it makes sense to switch from `joda` here as it is already used in some code examples and I see that our current requirements list `Java 1.8+`.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/utils/CodeSampleHelper.java
PierrickVoulet: Agreed, changed.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddAdCustomizer.java
```diff
@@ -141,7 +142,7 @@ private void runExample(GoogleAdsClient googleAdsClient, AddAdCustomizerParams p
           "Please pass exactly two ad group IDs in the adGroupId parameter.");
     }
 
-    String feedName = "Ad Customizer example feed " + System.currentTimeMillis();
+    String feedName = "Ad Customizer example feed " + CodeSampleHelper.getPrintableDatetime();
```
jradcliff, line 145: I'm inclined to have all these examples do a `static` import of `CodeSampleHelper.getPrintableDateTime()`.

@nwbirnie , any objections?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/utils/CodeSampleHelper.java
PierrickVoulet: ZonedDateTime was introduced in 1.8, switched to it.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddAdCustomizer.java
PierrickVoulet: I think it makes sense so made the change. This makes is much easier to read.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/advancedoperations/AddAdCustomizer.java
```diff
@@ -141,7 +143,7 @@ private void runExample(GoogleAdsClient googleAdsClient, AddAdCustomizerParams p
           "Please pass exactly two ad group IDs in the adGroupId parameter.");
     }
 
-    String feedName = "Ad Customizer example feed " + System.currentTimeMillis();
+    String feedName = "Ad Customizer example feed " + getPrintableDatetime();
```
jradcliff, line 146: @PierrickVoulet , could you trigger the functional tests against this PR to make sure that none of the names exceed product limits for an entity? With this change, we're going from around 13 characters (system millis) to 25.

PierrickVoulet: Good call, I few were failing due to size.  I added a shorter version of the printable data time (15 char-long) and used it for those that needed it.

PierrickVoulet: All functional tests pass, only the AddCampaignDraft fails but for unrelated issue.

jradcliff (approved):

nwbirnie (approved):

google-ads-examples/src/main/java/com/google/ads/googleads/examples/utils/CodeSampleHelper.java
```diff
@@ -0,0 +1,31 @@
...
+
+/** A helper class for all code examples. */
+public abstract class CodeSampleHelper {
+
+  /**
+   * Generates a printable string for the current date and time in UTC time zone.
+   *
+   * @return the result string.
+   */
+  public static String getPrintableDatetime() {
```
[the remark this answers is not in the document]
PierrickVoulet: I think `get` is better than `format` in this context, I guess we could easily change later on if this is confusing.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/utils/CodeSampleHelper.java
```diff
@@ -0,0 +1,48 @@
...
+  /** The shorter date format used for printing. */
+  private static final DateTimeFormatter shortFormat =
+      DateTimeFormatter.ofPattern("MMddHHmmssSSS");
+
+  /**
+   * Generates a printable string for the current date and time in local time zone.
+   *
+   * @return the result string.
+   */
+  public static String getPrintableDatetime() {
```
jradcliff, line 36: we should use the same capitalization as `ZonedDateTime`.

PierrickVoulet:
@jradcliff @nwbirnie I believe this PR is ready to merge. I cannot merge myself on this repo so I defer to you.

jradcliff (approved):

jradcliff:
Merged now. Thanks for making this improvement!

PR #386, opened 2020-12-09
Add support for configuring a GoogleAdsClient using environment variables

jradcliff:
Also added support for configuring `linkedCustomerId` from a config
file, properties, or an environment variable.


jradcliff:
Note to reviewers:

I considered multiple approaches here, but having a private `GoogleCredentials.Builder` was the cleanest approach I could find that satisfied all of the following requirements:

1. Support merging values from environment variables, a config file, properties, and setter invocations.
1. Do not make any breaking changes to the `public` interface of the client or its builder.
1. Support _either_ setting credentials explicitly via `setCredentials(Credentials)` _or_ configuring credentials from a file, properties, or environment variables.
1. Play nice with AutoValue. An unfortunate result of this requirement is that I had to mark `getCredentials()` on the client as `@Nullable`. There were alternatives to doing so, but they were all considerably less pleasant.

google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
```diff
@@ -154,39 +168,52 @@ public static Builder newBuilder() {
...
+            return new File(envConfigFilePath);
+          }
+          return new File(System.getProperty("user.home"), DEFAULT_PROPERTIES_CONFIG_FILE_NAME);
+        };
+
+    /** Allows injecting a different map of environment variables. Only modify this for testing. */
+    private Map<String, String> environment;
+
+    /**
+     * A builder for the client's Credentials. This will be not set, a UserCredentials.Builder, or a
```
devchas, line 185: Can you clarify this comment?

google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
```diff
@@ -154,39 +168,52 @@ public static Builder newBuilder() {
...
-                    + " for %s if using installed application/web flow credentials, or %s if using"
-                    + " service account credentials.",
-                INSTALLED_APP_OAUTH_KEYS, SERVICE_ACCOUNT_OAUTH_KEYS));
-      } else if (hasInstalledAppKeys && hasServiceAccountKeys) {
+        // Entries missing for both types of credentials. Skips any further processing so user can
+        // merge this builder if needed.
+        return;
+      }
+
+      if (hasInstalledAppKeys && hasServiceAccountKeys) {
```
devchas, line 216: This can be added to the prior if statement as an else condition.

google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
```diff
@@ -199,66 +226,113 @@ private void setCredentials(Properties properties) throws IOException {
...
+          throw new IllegalArgumentException(
+              "Entries found in properties for installed application/web flow credentials, but"
+                  + " this builder is already partially configured for service account"
+                  + " credentials.");
+        }
+      }
+
+      // Updates the credentials builder using the appropriate set of properties.
+      GoogleCredentials.Builder updatedBuilder;
+      if (hasInstalledAppKeys) {
```
devchas, line 247: Is this backward? If hasInstalledAppKeys, wouldn't this be a service account?

google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
```diff
@@ -408,6 +525,25 @@ public Builder fromProperties(Properties properties) {
...
+        }
+      } else {
+        // If the user both has credential values in the config file, properties, or environment
+        // variables, but they also explicitly set credentials via setCredentials(), their intent
+        // is not 100% clear. Instead of attempting to guess their intent by merging the former with
+        // the latter or always choosing one over the other (neither of which may be the user's
+        // intent), raise an exception and require the user to remove the ambiguity.
+        Preconditions.checkState(
+            !credentialsBuilder.isPresent(),
+            "Credentials were both explicitly set and configured from a file, properties, or "
```
devchas, line 543: Suggested edit to make this a little clearer:

```
            "Credentials were both (a) explicitly set via setCredentials() and (b) configured from a file, properties, or "
                + "environment variables. Either remove credentials settings from the latter or "
                + "do not call setCredentials().");
```

google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
jradcliff: Updated in the latest commit. PTAL.

jradcliff:
Heads up @nwbirnie and @devchas : I was able to find a way to implement "last one wins" logic for the 3rd requirement:

> 3. Support either setting credentials explicitly via setCredentials(Credentials) or configuring credentials from a file, properties, or environment variables.

With the latest commit, users who previously took the following steps in either order will see the same behavior as always:

* Call `setCredentials(Credentials)`.
* Call `fromProperties(...)`.
 
That is, the last step they performed will configure their credentials. My prior commits on this PR would have made either sequence throw an exception, which would have been a breaking change.

devchas (approved):

nwbirnie (commented):
First round of comments - going to take another pass just now

google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
```diff
@@ -199,66 +226,113 @@ private void setCredentials(Properties properties) throws IOException {
                 SERVICE_ACCOUNT_OAUTH_KEYS));
       }
 
-      // Constructs the credentials object using the appropriate set of properties.
-      setCredentials(
-          hasInstalledAppKeys
-              ? createUserCredentials(properties)
-              : createServiceAccountCredentials(properties));
+      if (credentialsBuilder.isPresent()) {
```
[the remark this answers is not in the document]
jradcliff: I did, but unfortunately `Credentials` does not have a builder. The builder inheritance chain stops at `OAuth2Credentials.Builder`.

```
Credentials  // No Builder here
  OAuth2Credentials // Builder here and below, with inheritance
    GoogleCredentials
      UserCredentials
      ServiceAccountCredentials
```

AutoValue nested builders require that `Credentials` has `Credentials.Builder()`, `Credentials.builder()`, or `Credentials.newBuilder()`.

I could work around this by declaring `getCredentials()` as returning a `GoogleCredentials` or an `OAuth2Credentials`, but that would be a breaking change. Same goes for if I introduced some wrapper around `Credentials`.

google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
```diff
@@ -199,66 +226,113 @@ private void setCredentials(Properties properties) throws IOException {
...
+      }
+
       String serviceAccountSecretsPath =
           ConfigPropertyKey.SERVICE_ACCOUNT_SECRETS_PATH.getPropertyValue(properties);
-      Preconditions.checkNotNull(
-          serviceAccountSecretsPath,
-          "No service account secrets path specified under the key: %s",
-          ConfigPropertyKey.SERVICE_ACCOUNT_SECRETS_PATH);
+      if (serviceAccountSecretsPath != null) {
+        // Unfortunately, ServiceAccountCredentials.Builder does not provide a fromStream() method
```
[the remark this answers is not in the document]
jradcliff: `fromStream()` populates at least 7 different attributes of the instance, as well as some other internals like the transport. I thought the safer choice here would be to let `fromStream()` do its thing, and then we can confidently copy the user and scopes since that covers everything we do with the `ServiceCredentials` interface without worrying about what's happening internally as a result of those calls.

I worry that if we dig into `fromStream()` and make sure we're copying everything it sets, this'll easily break if that implementation changes.

google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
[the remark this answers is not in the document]
nwbirnie: Glad we considered this. I'm not sure that an internal convenience is a good reason to introduce breaking changes to all users.

google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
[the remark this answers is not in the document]
nwbirnie: Good call, thanks for the info.

google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
```diff
@@ -442,6 +582,63 @@ Builder setConfigurationFileSupplier(Supplier<File> configurationFileSupplier) {
       return this;
     }
 
+    @VisibleForTesting
+    Builder setEnvironment(Map<String, String> environment) {
+      this.environment = ImmutableMap.copyOf(environment);
```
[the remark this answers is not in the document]
nwbirnie: This is a nit, feel free to choose how to resolve.

nwbirnie:
LGTM pending open comments

google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
```diff
@@ -154,39 +168,52 @@ public static Builder newBuilder() {
...
+            return new File(envConfigFilePath);
+          }
+          return new File(System.getProperty("user.home"), DEFAULT_PROPERTIES_CONFIG_FILE_NAME);
+        };
+
+    /** Allows injecting a different map of environment variables. Only modify this for testing. */
+    private Map<String, String> environment;
+
+    /**
+     * A builder for the client's Credentials. If this optional {@link Optional#isPresent()}, then
```
nwbirnie, line 185: nit: active voice

google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
```diff
@@ -95,10 +98,17 @@ public static Builder newBuilder() {
...
+    clientBuilder.setEnvironment(System.getenv());
     return clientBuilder;
   }
 
-  /** Returns the credentials for this client. */
+  /**
+   * Returns the credentials for this client.
+   *
+   * <p>This field is marked nullable to enable proper builder behavior. In practice, all requests
+   * to Google Ads API must be authenticated.
```
nwbirnie, line 109: nit: ... and this field can never be null.

(we should probably move away from AutoValue for this and similar reasons at some point.)

google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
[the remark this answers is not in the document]
jradcliff: Good suggestion. I swapped this out with a `Function<String, String>` and it worked out quite nicely + avoided copying the environment map.

google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
jradcliff: Updated wording.

jradcliff:
I merged from master at `d632e3ae` in preparation for merging this. Please use *Jump to* to look at the two files that actually changed: `GoogleAdsClient.java` and `GoogleAdsClientTest.java`.

nwbirnie (approved):

PR #387, opened 2020-12-09
add getFieldValue method to FieldMasks

devchas:
Change-Id: I1cebb3c4ec678164e3de4e47a1380d478aaba77d

### DIFF
--- a/google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
+++ b/google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
@@ -17,9 +17,13 @@ package com.google.ads.googleads.lib.utils;
 import com.google.common.base.Preconditions;
 import com.google.protobuf.Descriptors.Descriptor;
 import com.google.protobuf.Descriptors.FieldDescriptor;
+import com.google.protobuf.Descriptors.FieldDescriptor.JavaType;
 import com.google.protobuf.FieldMask;
 import com.google.protobuf.GeneratedMessageV3;
 import com.google.protobuf.Message;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Objects;
 
 /** Utility methods for working with field masks. */
@@ -117,6 +121,65 @@ public class FieldMasks {
         }
       }
     }
+  }
+
+  /**
+   * Gets the field value of a message from a field mask path.
+   *
+   * @param fieldMaskPath The field mask path.
+   * @param entity The entity to retrieve values from.
+   * @return the field Object.
+   */
+  public static Object getFieldValue(String fieldMaskPath, Message entity) {
+    List<String> fieldMaskParts = new ArrayList<>(Arrays.asList(fieldMaskPath.split("\\.")));
+    Message currentEntity = entity;
+    while (!fieldMaskParts.isEmpty()) {
+      String fieldMaskPart = fieldMaskParts.remove(0);
+      if (Objects.isNull(currentEntity)) {
+        throw new IllegalArgumentException(
+            String.format("Cannot get field value. %s is null.", fieldMaskPart));
+      }
+      Descriptor descriptor = currentEntity.getDescriptorForType();
+      FieldDescriptor childField = descriptor.findFieldByName(fieldMaskPart);
+      Object childValue;
+      try {
+        childValue = currentEntity.getField(childField);
+      } catch (NullPointerException e) {
+        throw new IllegalArgumentException(
+            String.format(
+                "Cannot retrieve field value. A matching field was not found after"
+                    + " navigating %s up to %s.",
+                fieldMaskPath, fieldMaskPart));
+      }
+      if (!childField.isRepeated()
+          && childField.getJavaType() == JavaType.MESSAGE
+          && childValue instanceof Message) {
+        // we can recurse.
+        currentEntity = (Message) childValue;
+      } else if (childField.isRepeated() && childValue instanceof List) {
+        if (fieldMaskParts.size() == 0) {
+          return childValue;
+        } else {
+          throw new IllegalArgumentException(
+              String.format(
+                  "Cannot retrieve field value. A repeated field was encountered after"
+                      + " navigating %s up to %s.",
+                  fieldMaskPath, fieldMaskPart));
+        }
+      } else {
+        if (fieldMaskParts.size() == 0) {
+          // we cannot recurse any longer.
+          return childValue;
+        } else {
+          throw new IllegalArgumentException(
+              String.format(
+                  "Cannot retrieve field value. A non-MESSAGE field was encountered after"
+                      + " navigating %s up to %s.",
+                  fieldMaskPath, fieldMaskPart));
+        }
+      }
+    }
+    return currentEntity;
   }
 
   private static String getFieldName(String currentField, FieldDescriptor field) {
--- a/google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
+++ b/google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
@@ -21,8 +21,9 @@ import com.google.protobuf.Descriptors.FieldDescriptor.JavaType;
 import com.google.protobuf.FieldMask;
 import com.google.protobuf.GeneratedMessageV3;
 import com.google.protobuf.Message;
-import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Deque;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
 
@@ -130,20 +131,20 @@ public class FieldMasks {
    * @param entity The entity to retrieve values from.
    * @return the field Object.
    */
-  public static Object getFieldValue(String fieldMaskPath, Message entity) {
-    List<String> fieldMaskParts = new ArrayList<>(Arrays.asList(fieldMaskPath.split("\\.")));
+  public static <V> V getFieldValue(String fieldMaskPath, Message entity) {
+    Deque<String> fieldMaskParts = new LinkedList<>(Arrays.asList(fieldMaskPath.split("\\.")));
     Message currentEntity = entity;
     while (!fieldMaskParts.isEmpty()) {
-      String fieldMaskPart = fieldMaskParts.remove(0);
+      String fieldMaskPart = fieldMaskParts.remove();
       if (Objects.isNull(currentEntity)) {
         throw new IllegalArgumentException(
             String.format("Cannot get field value. %s is null.", fieldMaskPart));
       }
       Descriptor descriptor = currentEntity.getDescriptorForType();
       FieldDescriptor childField = descriptor.findFieldByName(fieldMaskPart);
-      Object childValue;
+      V childValue;
       try {
-        childValue = currentEntity.getField(childField);
+        childValue = (V) currentEntity.getField(childField);
       } catch (NullPointerException e) {
         throw new IllegalArgumentException(
             String.format(
@@ -157,7 +158,7 @@ public class FieldMasks {
         // we can recurse.
         currentEntity = (Message) childValue;
       } else if (childField.isRepeated() && childValue instanceof List) {
-        if (fieldMaskParts.size() == 0) {
+        if (fieldMaskParts.isEmpty()) {
           return childValue;
         } else {
           throw new IllegalArgumentException(
@@ -167,7 +168,7 @@ public class FieldMasks {
                   fieldMaskPath, fieldMaskPart));
         }
       } else {
-        if (fieldMaskParts.size() == 0) {
+        if (fieldMaskParts.isEmpty()) {
           // we cannot recurse any longer.
           return childValue;
         } else {
@@ -179,7 +180,7 @@ public class FieldMasks {
         }
       }
     }
-    return currentEntity;
+    return (V) currentEntity;
   }
 
   private static String getFieldName(String currentField, FieldDescriptor field) {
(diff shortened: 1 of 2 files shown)


google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
```diff
@@ -119,6 +123,65 @@ private static void compare(
...
 
+  /**
+   * Gets the field value of a message from a field mask path.
+   *
+   * @param fieldMaskPath The field mask path.
+   * @param entity The entity to retrieve values from.
+   * @return the field Object.
+   */
+  public static Object getFieldValue(String fieldMaskPath, Message entity) {
+    List<String> fieldMaskParts = new ArrayList<>(Arrays.asList(fieldMaskPath.split("\\.")));
```
jradcliff, line 134: Removing the first element of an ArrayList can get expensive because it has to shift everything back a slot in the underlying array. It'd be more efficient to use a `LinkedList` here, and declare it as a `Deque`.

Then you can just do:

```
while (!fieldMaskParts.isEmpty()) {
   String fieldMaskPart = fieldMaskParts.remove();
   ...
}
```

google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
```diff
@@ -119,6 +123,65 @@ private static void compare(
...
+                    + " navigating %s up to %s.",
+                fieldMaskPath, fieldMaskPart));
+      }
+      if (!childField.isRepeated()
+          && childField.getJavaType() == JavaType.MESSAGE
+          && childValue instanceof Message) {
+        // we can recurse.
+        currentEntity = (Message) childValue;
+      } else if (childField.isRepeated() && childValue instanceof List) {
+        if (fieldMaskParts.size() == 0) {
```
jradcliff, line 160: nit: Use `fieldMaskParts.isEmpty()`

google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
```diff
@@ -119,6 +123,65 @@ private static void compare(
...
+          return childValue;
+        } else {
+          throw new IllegalArgumentException(
```
jradcliff, line 163: Up to you, but you could use `Preconditions.checkArgument` here and avoid the if/else.

google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
```diff
@@ -119,6 +123,65 @@ private static void compare(
...
+              String.format(
+                  "Cannot retrieve field value. A repeated field was encountered after"
+                      + " navigating %s up to %s.",
+                  fieldMaskPath, fieldMaskPart));
+        }
+      } else {
+        if (fieldMaskParts.size() == 0) {
```
jradcliff, line 170: `isEmpty()` here as well

google-ads/src/test/java/com/google/ads/googleads/lib/utils/FieldMasksTest.java
```diff
@@ -1,24 +1,102 @@
...
+            () -> {
+              FieldMasks.getFieldValue("url_custom_parameters.key", campaign);
+            });
+    assertEquals(
+        exception.getMessage(),
+        "Cannot retrieve field value. A repeated field was encountered after navigating"
+            + " url_custom_parameters.key up to url_custom_parameters.");
+  }
+
+  @Test
```
nwbirnie, line 90: Can you add a test case for attempting to access a subfield of a Message when the containing message is null.

nwbirnie (changes requested):

google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
```diff
@@ -119,6 +124,58 @@ private static void compare(
...
+    while (!fieldMaskParts.isEmpty()) {
+      String fieldMaskPart = fieldMaskParts.remove();
+      Preconditions.checkNotNull(
+          currentEntity, String.format("Cannot get field value. %s is null.", fieldMaskPart));
+      Descriptor descriptor = currentEntity.getDescriptorForType();
+      FieldDescriptor childField = descriptor.findFieldByName(fieldMaskPart);
+      V childValue;
+      try {
+        childValue = (V) currentEntity.getField(childField);
+      } catch (NullPointerException e) {
```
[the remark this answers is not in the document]
devchas: You'll never get to the if statement here because `descriptor.findFieldByName(fieldMaskPart)` throws a `NullPointerException` if `childField` is `null` which is why this is structured the way it is.

google-ads/src/test/java/com/google/ads/googleads/lib/utils/FieldMasksTest.java
devchas: I couldn't figure out a way to test this, but I'm open to ideas. There are very few cases where a value can actually be null. For example, [campaign.optimization_score](https://developers.google.com/google-ads/api/fields/v6/campaign#campaign.optimization_score) can be null, but this field is output only. Without digging through the entire API, it seems output only fields are the only field that could possibly have null values. With the `Campaign` created in the test, the optimization score defaults to 0.0. Moreover, you can't mock a resource on account of the fact that resources are `final`. That's the reason for the missing test coverage here, but I'm open to ideas.

google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
```diff
@@ -119,6 +124,58 @@ private static void compare(
...
+        throw new IllegalArgumentException(
```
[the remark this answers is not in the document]
devchas: I don't think that makes sense here. While we are parsing the path, it isn't an error because we couldn't parse. The error results from the fact that a user entered an invalid argument that fails to meet some specified condition.

google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
```diff
@@ -119,6 +124,58 @@ private static void compare(
...
   }
 
+  /**
+   * Gets the field value of a message from a field mask path.
+   *
+   * @param fieldMaskPath The field mask path.
+   * @param entity The entity to retrieve values from.
+   * @return the field Object.
+   */
+  public static <V> V getFieldValue(String fieldMaskPath, Message entity) {
```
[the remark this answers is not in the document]
devchas: This doesn't work here because of the case where an enum value is returned, in which case, an optional cannot be returned.

google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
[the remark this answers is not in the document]
nwbirnie: Hmm, I'd still prefer not to catch `NullPointerException`. This could be caused by any number of things, not just because the field we passed was null. Can we just allow this block to throw?

google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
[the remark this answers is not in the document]
nwbirnie: Screenshot from 2020-12-11 14-57-45

Looks like findFieldByName() returns null if the field doesn't exist.

google-ads/src/test/java/com/google/ads/googleads/lib/utils/FieldMasksTest.java
nwbirnie: That's a good point, in proto3 even message fields are never null.

google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
```diff
@@ -119,6 +124,49 @@ private static void compare(
...
+                    + " navigating %s up to %s",
+                fieldMaskPath, fieldName));
+      }
+
+      if (fieldMaskParts.isEmpty()) {
+        return childValue;
+      } else if (childValue == null) {
+        throw new IllegalArgumentException("Attempt to access a sub-field of " + fieldName + " which is null");
+      } else if (childField.isRepeated()) {
+        throw new IllegalArgumentException("Cannot access repeated sub-field " + fieldName + " of " + currentEntity.getClass());
```
nwbirnie, line 160: Might be helpful here to include the `fieldMaskPath` and `fieldName` as well.

google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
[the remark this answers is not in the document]
nwbirnie: Looks like there's 2 exception cases we need to handle

- The field doesn't exist (the case this comment is tagged on).
- Attempting to access an element of a repeated/primitive/enum field.

Checking the [style guide](http://go/java-practices/exceptions#checked_unchecked), the checked exception should only be used if both:

- the exception can be avoided by proper use of the API
- and the programmer can take some useful action

I'm actually not sure that either condition holds, so unchecked is actually the right call.

For the first case, if the field doesn't exist, it looks like we can detect this because `getFieldByName()` returns null. In this case, rather than triggering the null pointer exception on `getValue()` it would be better to check if the field is null and throw our own exception. This (1) makes it clear what the failure modes are, (2) avoids catching potentially unrelated exceptions and rewrapping them with the wrong message, and (3) avoids additional cost of throwing 2 exceptions.

nwbirnie (changes requested):

nwbirnie (changes requested):

google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
```diff
@@ -119,6 +124,57 @@ private static void compare(
...
+        new LinkedList<>(Arrays.asList(fieldMaskPath.trim().split("\\.")));
+    Message currentEntity = entity;
+    while (!fieldMaskParts.isEmpty()) {
+      String fieldName = fieldMaskParts.remove();
+      Preconditions.checkNotNull(
+          currentEntity, String.format("Cannot get field value. %s is null", fieldName));
+      Descriptor descriptor = currentEntity.getDescriptorForType();
+      FieldDescriptor childField = descriptor.findFieldByName(fieldName);
+
+      V childValue;
```
nwbirnie, line 145: No need to separate the declaration from assignment here.

nwbirnie (approved):

google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
```diff
@@ -119,6 +124,56 @@ private static void compare(
...
+      FieldDescriptor childField = descriptor.findFieldByName(fieldName);
+
+      if (childField == null) {
+        throw new IllegalArgumentException(
+            String.format(
+                "Cannot retrieve field value. A matching field was not found after"
+                    + " navigating %s up to %s",
+                fieldMaskPath, fieldName));
+      }
+      V childValue = (V) currentEntity.getField(childField);
```
jradcliff, line 152: Will this cast fail on intermediate fields, since `V` is the type of the leaf field, not all the fields leading up to it?
Instead, you could just keep the casts you have on lines 154 and 174.

google-ads/src/test/java/com/google/ads/googleads/lib/utils/FieldMasksTest.java
```diff
@@ -1,24 +1,101 @@
...
+  public void getFieldValue_gets_top_nested_field() {
+    assertEquals(FieldMasks.getFieldValue("manual_cpc", campaign), campaign.getManualCpc());
+  }
+
+  @Test
+  public void getFieldValue_gets_enum_field() {
+    assertEquals(
```
jradcliff, line 56: I noticed none of the tests attempts to assign the result of `FieldMasks.getFieldValue()` to a specific type (e.g., `CampaignStatusEnum.CampaignStatus` here). Could you add that to make sure the casts work as expected?

google-ads/src/main/java/com/google/ads/googleads/lib/utils/FieldMasks.java
devchas: This seems to work for all fields in the path and won't compile without it, so I don't think we need to keep as is.

jradcliff: I'm not sure if you'll see exceptions if you add tests as I mentioned in my [other comment](https://github.com/googleads/google-ads-java/pull/387#discussion_r542666622). Type erasure is weird in Java, so it's possible it'll still work, but I'll leave this open as a reminder to check after you address the other comment.

google-ads/src/test/java/com/google/ads/googleads/lib/utils/FieldMasksTest.java
devchas: I've added tests here to demonstrate all cases of assignment. The only "quirk" is that you cannot assign an enum value directly. You have to assign by number because the value returned is an `EnumValueDescriptor`. Even with the casting change you mentioned in your other comment, which I've made, you still have to approach it in this way.

jradcliff: Thanks, and I see now what you mean about enums. Perhaps a future enhancement would be to return the value as the typed enum, but there are edge cases we should consider first (like unknown values), so I won't hold up this enhancement for that.

jradcliff (approved):

PR #392, opened 2020-12-11
Update primer files for v6 and GAX version change

jradcliff:


nwbirnie (approved):

google-ads/src/main/resources/googleadsapi-primer/protobuf_class_names
```diff
@@ -381,3 +381,1298 @@ com.google.ads.googleads.v5.services.UserInterestServiceClient
 com.google.ads.googleads.v5.services.UserListServiceClient
 com.google.ads.googleads.v5.services.UserLocationViewServiceClient
 com.google.ads.googleads.v5.services.VideoServiceClient
```
nwbirnie, line 383: Can remove v5 classes.

google-ads/src/main/resources/googleadsapi-primer/grpc_class_names
```diff
@@ -1,155 +1,363 @@
-io.grpc.Attributes$Builder
+io.grpc.alts.ComputeEngineChannelBuilder
 io.grpc.Attributes
+io.grpc.Attributes$Builder
+io.grpc.Attributes$Key
+io.grpc.auth.GoogleAuthLibraryCallCredentials
 io.grpc.auth.GoogleAuthLibraryCallCredentials$1
```
nwbirnie, line 6: Need to remove the nested classes - Class.forName() doesn't load these so it slows everything down.

nwbirnie (changes requested):

google-ads/src/main/resources/googleadsapi-primer/grpc_class_names
nwbirnie: Actually maybe I'm misremembering... I can see that we originally had some in here.

nwbirnie (approved):

PR #411, opened 2021-03-22
Use getPrintableDateTime for name in AddCustomAudience

jradcliff:
Fixes the one remaining `System.currentTimeMillis()` usage in `google-ads-examples`.


nwbirnie (approved):

nwbirnie:
Thanks for fixing this!

Issue #412, opened 2021-03-23
Issue Creating Campaign due to failure during CampaignBudget Creation.


armelHx:
Hello team, 

We're getting errors while creatingCampaigns due to the following error below: 

```2021-03-19 14:02:21,704 WARN  [TRACEID:  ,SPANID: ] [Gax-832] [] c.g.ads.googleads.lib.request.summary   : FAILURE REQUEST SUMMARY. Method: google.ads.googleads.v6.services.CampaignBudgetService/MutateCampaignBudgets, Endpoint: googleads.googleapis.com:443, CustomerID: 8497610439, RequestID: zorax6u-iwt-gzgBXQCL5A, ResponseCode: INVALID_ARGUMENT, Fault: Request contains an invalid argument..
2021-03-19 14:02:21,704 INFO  [TRACEID:  ,SPANID: ] [Gax-832] [] c.g.ads.googleads.lib.request.detail    : FAILURE REQUEST DETAIL.
Request
-------
MethodName: google.ads.googleads.v6.services.CampaignBudgetService/MutateCampaignBudgets
Endpoint: googleads.googleapis.com:443
Headers: {developer-token=REDACTED, login-customer-id=5531825568, x-goog-api-client=gl-java/11.0.3 gapic/ gax/1.57.0 grpc/1.30.2}
Body: [FAILED toString()]

Response
--------
Headers: Metadata(content-type=application/grpc,request-id=zorax6u-iwt-gzgBXQCL5A,date=Fri, 19 Mar 2021 14:02:21 GMT,alt-svc=h3-29=":443"; ma=2592000,h3-T051=":443"; ma=2592000,h3-Q050=":443"; ma=2592000,h3-Q046=":443"; ma=2592000,h3-Q043=":443"; ma=2592000,quic=":443"; ma=2592000; v="46,43")
Body: null
Failure message: [FAILED toString()]
Status: Status{code=INVALID_ARGUMENT, description=Request contains an invalid argument., cause=null}.
2021-03-19 14:02:21,707 ERROR [TRACEID:  ,SPANID: ] [Gax-832] [] c.g.c.util.concurrent.AbstractFuture    : RuntimeException while executing runnable CallbackListener{com.google.api.core.ApiFutures$1@67a0a9b9} with executor MoreExecutors.directExecutor()
java.lang.RuntimeException: Generated message class "com.google.ads.googleads.v6.errors.ErrorLocation$FieldPathElement" missing method "getIndexCase".
```

[...]

PR #413, opened 2021-03-23
Add protobuf to requirements.

nwbirnie:
Idea is to avoid issues like #412 in future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ This project hosts the Java client library for the Google Ads API.
   * Maven 3.0+
 
 ## Compatible dependencies
+
   * [Protobuf](https://github.com/protocolbuffers/protobuf) 3.12.2+
   * [Guava](https://github.com/google/guava) 30.0+
 
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ This project hosts the Java client library for the Google Ads API.
 
   * Java 1.8+
   * Maven 3.0+
+  * Protobuf >= 3.12.2
 
 ## Maven artifacts
 
--- a/README.md
+++ b/README.md
@@ -12,7 +12,10 @@ This project hosts the Java client library for the Google Ads API.
 
   * Java 1.8+
   * Maven 3.0+
+
+## Compatible dependencies
   * [Protobuf](https://github.com/protocolbuffers/protobuf) 3.12.2+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,6 +12,7 @@ This project hosts the Java client library for the Google Ads API.
 
   * Java 1.8+
   * Maven 3.0+
+  * Protobuf 3.12.2+
```
jradcliff, line 15: Are you also seeing gRPC dependencies as an issue, or is protobuf by far the most common problem?

jradcliff, line 15: Suggest making "Protobuf" a link to the protobuf library

jradcliff, line 15: How frequently do you anticipate this version will need to change? Wondering if it makes sense to automate keeping this in sync with poms, like you do with the library version on line 22.

nwbirnie: I'm not aware of a specific lower bound for gRPC. guava is a more common source of issues. I'll add this to the line up as well.

nwbirnie: Done

nwbirnie: Good point, I'll add this to the release tasks for the next release.

devchas (approved):

README.md
jradcliff: I'm not trying to complicate things (honest!), but would it be better if this was a separate section, or a "Dependencies" subsection of "Requirements"?

The protobuf and guava dependency requirements are different from the Java runtime requirement. The former only come into play if the user _already_ includes protobuf or guava as a transitive or direct dependency in their project. If not, google-ads-java will work just fine. In contrast, if the user tries to use Java 1.5 :tiger: , it's likely they'll run into issues.

Perhaps something like:

```
## Compatible dependencies

*  protobuf 3.12.2+
*  guava x.y.z+
```

nwbirnie: Sounds good. Updated.

README.md
```diff
@@ -13,6 +13,10 @@ This project hosts the Java client library for the Google Ads API.
   * Java 1.8+
   * Maven 3.0+
 
+## Compatible dependencies
+  * [Protobuf](https://github.com/protocolbuffers/protobuf) 3.12.2+
```
jradcliff, line 17: nit: add a blank line after the header to match style in rest of the doc

jradcliff (approved):
One tiny nit, else LGTM

PR #428, opened 2021-05-05
Bump GAX to 1.63.0

### LINKED ISSUE #359: Difficulty setting up retry settings with GoogleAdsClient
Kotlin 1.4.10, jvmTarget 11

Linux, Ubuntu 18.04

 We're migrating one of our reports from ruby to java (well, ok...kotlin).

In implementing this,  I ran into a server error, which in ruby we'd retry.

Then, as I tried to set up retries via the java client, I ran into some weird inconsistencies.

Originally, I had (abbreviate):

```
        val gClient = GoogleAdsClient
                .newBuilder()
                .setCredentials(creds)
                .setDeveloperToken(devToken)
                .setLoginCustomerId(CG_MCC_ID)
                .build()

       val totalInserts = gClient.latestVersion.createGoogleAdsServiceClient().use { gServClient ->
```

etc.

However, the client I get from `GoogleAdsClient.latestVersion.createGoogleAdsServiceClient` did not appear to have a way to modify the `GoogleAdsServiceSettings`

So I ended up doing

`GoogleAdsServiceClient.create(getDefaultRetrySettings(creds, headerProvider))`  (where I create a settings builder and set everything in the `getDefaultRetrysettings` function)

But in order to do the latter, I had to implement a custom `HeaderProvider` to set the developer token and the login customer id.

Did I miss something in one of those two versions?

I'd expect to either be able to mutate the retry settings in the former, or to be able to set developer token and login-customer-id in the latter.

Thanks!

sean

nwbirnie:
Closes #359.

Retry settings are now available in GrpcClientContext.


nwbirnie:
Closes #359.

Retry settings are now available in GrpcClientContext.

devchas (approved):

devchas (changes requested):
It seems the gax bump is changing the effective dependency tree, so we may now have to explicitly declare commons-lang3 in the examples project, which is the reason for the kokoro failure.

google-ads/pom.xml
```diff
@@ -36,7 +36,7 @@
     <auto-value.version>1.7.3</auto-value.version>
     <!-- Keep all versions below consistent with the io.netty version. -->
     <netty.version>2.0.26.Final</netty.version>
-    <gax.version>1.60.1</gax.version>
+    <gax.version>1.63.0</gax.version>
```
jradcliff, line 39: Was 1.60.1 used for latency testing? If so, this change will require another round of tests before release.

nwbirnie: Ah, yeah you're right. Will get a fresh set of numbers on this before releasing.

nwbirnie: Planning to merge this CL to master and use the master build (already verified that this is up and running).

devchas (approved):

google-ads/pom.xml
jradcliff: Sounds good. Thanks.

jradcliff (approved):

PR #445, opened 2021-06-18
Add sonatype publish for shadow jar

nwbirnie:
Change-Id: Iab14eeda17b2f8f79d80d78de97bfc6d365958da

### DIFF
--- a/google-ads/build.gradle
+++ b/google-ads/build.gradle
@@ -181,9 +181,9 @@ task copyLicenses(type: Copy) {
     // with legal obligations.
     if (tasks.generateLicenses.licenseFiles.size() != inputs.sourceFiles.size()) {
       throw new GradleException(
-              "Failed to locate files to copy from. Expected: "
-                      + tasks.generateLicenses.licenseFiles
-                      + " but was: " + inputs.sourceFiles.toList())
+          "Failed to locate files to copy from. Expected: "
+              + tasks.generateLicenses.licenseFiles
+              + " but was: " + inputs.sourceFiles.toList())
     }
   }
 }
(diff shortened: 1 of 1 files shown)


google-ads/build.gradle
```diff
@@ -275,3 +277,79 @@ task testShadowJar {
...
+      def credentialsConfig = file(System.properties['user.home'] + '/.sonatype-creds')
+      if (credentialsConfig.exists()) {
+        def props = new Properties()
+        credentialsConfig.withInputStream { props.load(it) }
+        credentials {
+          username props['username']
+          password props['password']
+        }
+      } else {
+        logger.warn("Unable to load Sonatype credentials from ~/.sonatype-creds. Publishing will fail.")
```
jradcliff, line 351: I like having this warning, but I'm a little concerned about showing it to every user who runs any gradle task.

Can you instead move this to the first step of publishing? It could even be an error for that case.

nwbirnie: Added a check.

jradcliff: Thanks. Confirmed I no longer get the warning for non-pub tasks.

jradcliff (approved):

AndrewMBurke (approved):

google-ads/build.gradle
```diff
@@ -275,3 +277,90 @@ task testShadowJar {
...
+        id = 'josh'
+        name = 'Josh Radcliff'
+        email = 'jradcliff@users.noreply.github.com'
+        organization = 'Google'
+        organizationUrl = 'https://www.google.com'
+      }
+      developer {
+        id = 'nick'
+        name = 'Nick Birnie'
+        email = 'jradcliff@users.noreply.github.com'
```
devchas, line 308: Josh's email is used for your and my entries.

devchas (approved):
LGTM with minor fix.

devchas (approved):

PR #450, opened 2021-06-25
NEW: Add ability to change created GoogleAdsClient (Fixes #174)

### LINKED ISSUE #174: LoginCustomerId is not reset after GoogleAdsClient.toBuilder().setLoginCustomerId()

devchas:
This PR replaces #430.

Some notes:
- I don't think we have to worry about clobbering custom `TransportChannelProvider` settings. The reason for this is that we do not expose an interface for updating the `TransportChannelProvider` outside of the package. Therefore, if someone is creating a custom `TransportChannelProvider`, they must be doing so by customizing the `GoogleAdsClient`, in which case, they'll need to update this implementation anyhow.
- I have tested that the logging interceptor is updated to reflect the new `loginCustomerId` manually, but I don't think it is possible to write a test for this for two reasons. First, we don't own the `TransportChannelProvider` class, so we aren't able to write a method that exposes the interceptor provider, even with a VisibleForTesting annotation. Second, the interceptor provider is not ever set in the tests because they use `LocalChannelProvider`.
- The reason for my implementation is that the only case that breaks our tests is when the `TransportChannelProvider` is a `LocalChannelProvider` (i.e. all of our tests). I can't use `instance of` here because `LocalChannelProvider` is only available in the testing package.

### DIFF
--- a/google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
+++ b/google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
@@ -552,9 +552,7 @@ public abstract class GoogleAdsClient extends AbstractGoogleAdsClient {
       Primer.getInstance().ifPresent(p -> p.primeCredentialsAsync(getCredentials()));
       // Proceeds with creating the client library instance.
       TransportChannelProvider transportChannelProvider = getTransportChannelProvider();
-      if (transportChannelProvider.needsHeaders()) {
-        transportChannelProvider = transportChannelProvider.withHeaders(getHeaders());
-      }
+      transportChannelProvider = transportChannelProvider.withHeaders(getHeaders());
       if (transportChannelProvider.needsEndpoint()) {
         transportChannelProvider = transportChannelProvider.withEndpoint(getEndpoint());
       }
--- a/google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
+++ b/google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
@@ -551,8 +551,8 @@ public abstract class GoogleAdsClient extends AbstractGoogleAdsClient {
       // Provides the credentials to the primer to preemptively get these ready for usage.
       Primer.getInstance().ifPresent(p -> p.primeCredentialsAsync(getCredentials()));
       // Proceeds with creating the client library instance.
-      TransportChannelProvider transportChannelProvider = getTransportChannelProvider();
-      transportChannelProvider = transportChannelProvider.withHeaders(getHeaders());
+      TransportChannelProvider transportChannelProvider =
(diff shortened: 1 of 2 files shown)


google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
```diff
@@ -550,15 +550,40 @@ public GoogleAdsClient build() {
...
       // Provides the credentials to the primer to preemptively get these ready for usage.
       Primer.getInstance().ifPresent(p -> p.primeCredentialsAsync(getCredentials()));
-      // Proceeds with creating the client library instance.
-      TransportChannelProvider transportChannelProvider = getTransportChannelProvider();
-      if (transportChannelProvider.needsHeaders()) {
-        transportChannelProvider = transportChannelProvider.withHeaders(getHeaders());
-      }
-      if (transportChannelProvider.needsEndpoint()) {
-        transportChannelProvider = transportChannelProvider.withEndpoint(getEndpoint());
+      // Checks if the TransportChannelProvider is of type LocalChannelProvider in which case we
```
nwbirnie, line 553: The only time we can cast to InstantiatingGrpcChannelProvider is when it's a InstantiatingGrpcChannelProvider, so why not check for that?

google-ads/src/main/java/com/google/ads/googleads/lib/GoogleAdsClient.java
```diff
@@ -550,15 +550,40 @@ public GoogleAdsClient build() {
...
+        transportChannelProvider =
+            transportChannelProvider.toBuilder()
+                .setInterceptorProvider(
+                    () ->
+                        ImmutableList.of(
+                            new LoggingInterceptor(
+                                new RequestLogger(), getHeaders(), getEndpoint())))
+                .build();
+        // Proceeds with creating the client library instance.
+        transportChannelProvider =
```
nwbirnie, line 578: What if the user called setTransportChannelProvider(x) where x is InstantiatingGrpcChannelProvider?

I think the approach where we don't call setTransportChannelProvider seems more logical. Instead, we check if getTransportChannelProvider() == null and if so then we create an InstantiatingGrpcChannelProvider. Is there some reason not to do that here?

devchas: You cannot check if getTransportChannelProvider() == null. There is null precondition on getTransportChannelProvider(), so if it is null, an error is thrown.

nwbirnie: You should be able to add @Nullable to the method signature if that's a problem.

devchas: To which method signature? TransportChannelProvider? We don't own that class.

PR #473, opened 2021-08-23
Bump version for shadow jar

nwbirnie:
Change-Id: I93c57f4d1976e568a92e89feafa5248f982cd2b3

### DIFF
--- a/buildSrc/src/main/groovy/com.google.api-ads.java-conventions.gradle
+++ b/buildSrc/src/main/groovy/com.google.api-ads.java-conventions.gradle
@@ -35,7 +35,7 @@ repositories {
 }
 
 group = 'com.google.api-ads'
-version = '14.0.1-SNAPSHOT'
+version = '15.0.0'
 java.sourceCompatibility = JavaVersion.VERSION_1_8
 java.targetCompatibility = JavaVersion.VERSION_1_8
 
--- a/google-ads/build.gradle
+++ b/google-ads/build.gradle
@@ -369,7 +369,7 @@ publishing {
 
       groupId = 'com.google.api-ads'
       artifactId = 'google-ads-shadowjar'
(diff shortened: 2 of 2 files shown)


google-ads/build.gradle
```diff
@@ -369,7 +369,7 @@ publishing {
 
       groupId = 'com.google.api-ads'
       artifactId = 'google-ads-shadowjar'
-      version = '0.0.1-alpha'
+      version = '1.0.0'
```
jradcliff, line 372: If this is ready to move out of the alpha phase, should the version match the overall library version? I'm concerned another versioning scheme could be confusing on top of the library and API versions.

nwbirnie: Good call, switched this over to use the module-wide versioning

buildSrc/src/main/groovy/com.google.api-ads.java-conventions.gradle
```diff
@@ -35,7 +35,7 @@ repositories {
 }
 
 group = 'com.google.api-ads'
-version = '14.0.1-SNAPSHOT'
+version = '15.0.0'
```
jradcliff, line 38: Thanks for centralizing this. It should be `15.0.1-SNAPSHOT` to match the pom though, correct?

nwbirnie: Well the release that just went live is 15.0.0, and this change was intended to be used to release the shadow jar. You are correct though that this should be 15.0.1-SNAPSHOT. I could complete the release and bump again?

jradcliff: That SGTM.

jradcliff (approved):

devchas (approved):

PR #497, opened 2021-11-05
Add task to bump versions in gradle

nwbirnie:


build.gradle
```diff
@@ -0,0 +1,47 @@
+import java.util.regex.Pattern
+
+/** Defines a custom task for updating version numbers. */
+abstract class UpdateVersionTask extends DefaultTask {
+
+    static final def VERSION_REGEX = "([0-9]+)\\.([0-9]+)\\.([0-9]+)(-SNAPSHOT)?"
```
jradcliff, line 6: Add `^version=` to make sure this doesn't pick up any other lines.

nwbirnie: We're parsing into a `java.util.Properties`, so there's no need to match the raw text here.

jradcliff: Oh, of course! Sorry, didn't think of that.

jradcliff (approved):

devchas (approved):

Issue #507, opened 2021-11-22
JDK 15 as required in 16.0.0 and 15.0.1-SNAPSHOT release version for 16.


NickKononovich:
Hi team. I checked tags and seems 16.0.0 was released with 15.0.1-SNAPSHOT of Google Ads version.
And also and most important, I have JDK 11 and while building the project I have - java.lang.IllegalArgumentException: Unsupported class file major version 59. Could you please tell if 16.0.0 is supported starting from JDK 15? Thanks.

jradcliff:
Hi Nick,

Do you have the same problems if you use Gradle instead of Maven? I realize there are some issues with the poms as released, but hopefully using Gradle will work for you while we fix this.

Also, if you only want to use the library as a dependency, you shouldn't have to compile it locally. Setting your dependency on our library to `com.google.api-ads:google-ads:16.0.0` should suffice.

It appears the `.class` files in the JARs are for Java 1.8 (as expected), not JDK15. For example, I downloaded google-ads-16.0.0.jar from the central Maven repo, extracted it, and checked using `files`:

```
$ file com/google/ads/googleads/v7/utils/ResourceNames.class
com/google/ads/googleads/v7/utils/ResourceNames.class: compiled Java class data, version 52.0 (Java 1.8)
```

Thanks,
Josh

NickKononovich:
Hi Josh, Thank you for feedback. 
If I check version of this file 
`javap -verbose META-INF/versions/15/org/bouncycastle/jcajce/provider/asymmetric/edec/BC11XDHPrivateKey.class | grep "major"`
  major version: 59
Could you please check this files?

nwbirnie:
I tried to reproduce the issue by creating a new maven project, depending on `google-ads` and creating a new instance of `GoogleAdsClient`. I tried compiling this with JDK 8 and JDK 11, but [didn't get any compiler errors](https://gist.github.com/nwbirnie/a95f9692c2b6aa486bf7ac2562a86dad).

I also looked at the jar containing `META-INF/versions/15/org/bouncycastle/jcajce/provider/asymmetric/edec/BC11XDHPrivateKey.class`. This is found in [`bcprov-jdk15on-1.67.jar`](https://github.com/googleads/google-ads-java/blob/main/third_party/bcprov-jdk15on-1.67.jar). This is a dependency of gRPC, specifically `grpc-xds`.

[...]

PR #510, opened 2021-11-23
Switch grpc dependency to minimal required

nwbirnie:
#507 suggests that we should use the minimum set of grpc dependencies, rather than grpc-all

### DIFF
--- a/buildSrc/src/main/groovy/com.google.api-ads.java-conventions.gradle
+++ b/buildSrc/src/main/groovy/com.google.api-ads.java-conventions.gradle
@@ -153,7 +153,9 @@ dependencies {
   implementation 'com.google.api:gax'
   implementation 'com.google.api:gax-grpc'
   implementation 'com.google.protobuf:protobuf-java'
-  implementation 'io.grpc:grpc-all'
+  implementation 'io.grpc:grpc-stub'
+  implementation 'io.grpc:grpc-protobuf'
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/groovy/com.google.api-ads.java-conventions.gradle
```diff
@@ -153,7 +153,9 @@ dependencies {
   implementation 'com.google.api:gax'
   implementation 'com.google.api:gax-grpc'
   implementation 'com.google.protobuf:protobuf-java'
-  implementation 'io.grpc:grpc-all'
+  implementation 'io.grpc:grpc-stub'
```
jradcliff, line 156: Does `google-cloud-shared-dependencies` actually already cover these?

https://github.com/googleapis/java-shared-dependencies/blob/main/first-party-dependencies/pom.xml

nwbirnie: Yep, it imports [gRPC BOM](https://github.com/googleapis/java-shared-dependencies/blob/37395d148ab5cb4ccf9bbe05e21656e81ec53a89/first-party-dependencies/pom.xml#L75-L80) which declares [grpc-protobuf and grpc-stub](https://repo1.maven.org/maven2/io/grpc/grpc-bom/1.42.1/grpc-bom-1.42.1.pom).

jradcliff (approved):

PR #527, opened 2021-12-10
Move shadow jar to its own project.

### LINKED ISSUE #525: External dependencies are included in the jar
**Describe the bug:**
Libraries like Guava are included in com.google.api-ads:google-ads:16.0.1. That makes the management of these libraries impossible for us when using them in our application too. That bug has been introduced with version 16.0.0. Version 15.0.0
looks fine.

**Steps to Reproduce:**
Look into the jar. It includes e.g. com.google.common.collect.Maps which is part of Guava. There are many many more examples like e.g.  apache commons, slf4j, netty, ....

**Expected behavior:**
Declare these dependencies only as transitive dependencies or include them as shaded jars.

**Client library version and API version:**
Client library version: 16.0.1
Google Ads API version: v9
JDK version: 17

nwbirnie:
Fixes #525

Change-Id: I60b9e51010737b510f4071823ea9e409efd0881d


google-ads-shadowjar/build.gradle
```diff
@@ -0,0 +1,98 @@
...
+
+            // Allows the smoke tests to pass unless we're publishing to sonatype.
+            ignoreExitValue = !project.properties.containsKey("release")
+        }
+    }
+}
+
+publishing {
+    publications {
+        shadow(MavenPublication) { publication ->
```
jradcliff, line 83: I'm getting the following from a `publishToMavenLocal`. Is that simply because I've published locally before, or does it indicate an issue with the build configuration?

```
> Task :google-ads-shadowjar:publishShadowPublicationToMavenLocal
Multiple publications with coordinates 'com.google.api-ads:google-ads-shadowjar:16.0.1-SNAPSHOT' are published to repository 'mavenLocal'. The publications will overwrite each other!
```

nwbirnie: Good point, this was because the shadow plugin automatically creates a publication.

jradcliff: Thanks. Just tried again with this fix and didn't get the warning.

jradcliff (approved):

PR #531, opened 2021-12-10
Modify SmokeTest so it passes for manager accounts

jradcliff:
Fixes two issues:

1. Clears the login customer ID in case ads.properties happens to
   set it. This avoids errors if the credentials grant access to
   accounts in different hierarchies.
2. Queries customers instead of campaigns so queries return results
   even if the credentials only have access to manager accounts.

### DIFF
--- a/google-ads/src/test/functional/java/com/google/ads/googleads/lib/SmokeTest.java
+++ b/google-ads/src/test/functional/java/com/google/ads/googleads/lib/SmokeTest.java
@@ -41,7 +41,7 @@ import org.threeten.bp.Duration;
  *
  * <p>Only depends on a basic configuration file with devtoken, OAuth etc.
  *
- * <p>Retrieves all accessible customers and iterates through all their campaigns.
+ * <p>Retrieves all accessible customers and issues a search request for each customer.
  */
 @RunWith(JUnit4.class)
 public class SmokeTest {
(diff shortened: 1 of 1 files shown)


google-ads/src/test/functional/java/com/google/ads/googleads/lib/SmokeTest.java
```diff
@@ -76,22 +86,22 @@ public void ensureCanReadAllAvailableCustomersCampaigns() throws IOException {
...
-                    + "WHERE campaign.status != 'REMOVED'");
-        numCampaigns += StreamSupport.stream(response.iterateAll().spliterator(), false).count();
+                "SELECT customer.id, customer.manager, customer.currency_code,"
+                    + " customer.descriptive_name FROM customer");
+        numCustomers += StreamSupport.stream(response.iterateAll().spliterator(), false).count();
       }
     }
-    System.out.println("Retrieved " + numCampaigns + " campaigns");
-    assertTrue("Expected to read at least one campaign for smoke test.", numCampaigns > 0);
+    System.out.println("Retrieved " + numCustomers + " customers");
```
nwbirnie, line 104: I guess reading campaigns was slow?

nwbirnie (approved):

google-ads/src/test/functional/java/com/google/ads/googleads/lib/SmokeTest.java
jradcliff: Not quite. It was that if the accessible accounts consists solely of manager accounts, then the query won't find any campaigns and the test will fail. I think querying `customer` is an equally valid test, since it still exercises the library on a `search` request.

PR #538, opened 2021-12-13
Fixup publishing.

nwbirnie:
Avoids spurious warnings about losing testFixtures from maven publication.
Also declares BOM as API dependency so that dependents can read this.

Change-Id: Iba153ab3b40bfde7f6b4f7dbbe8ae96440833e14


jradcliff (approved):

nwbirnie:
This is only required for the modules refactor merge in #530.

PR #544, opened 2021-12-23
Implements a client for Sonatype Nexus

nwbirnie:
By default when you deploy a release to Nexus, it uses a feature known as auto-staging. This uses a heuristic to determine how subsequent PUT requests are aggregated into common staging repositories. Invariably Gradle's publish plugin confuses this, possibly because it creates new TCP connections for every artifact (you can see this when running with --debug), or maybe because the Nexus server is overloaded. Rather than coerce the children into playing nice, this change creates a plugin which creates a named staging repository and deploys the release into it.

The plugin functionality boils down to:

1. Discover the staging profiles on Nexus with a GET request.
1. Create a new staging repository with a POST request.
2. Override the nexus URL given to the maven-publish plugin with https://oss.sonatype.org/service/local/staging/deployByRepositoryId/<the repo ID we just created>.

There's more detailed documentation in the code. The implementation was based on these docs: https://support.sonatype.com/hc/en-us/articles/213465868-Uploading-to-a-Staging-Repository-via-REST-API

Change-Id: I19767b3ccd99461b167997ea15730f109dff9443

### DIFF
--- a/buildSrc/src/main/groovy/com.google.api-ads.nexus-publish.gradle
+++ b/buildSrc/src/main/groovy/com.google.api-ads.nexus-publish.gradle
@@ -134,7 +134,8 @@ class SonatypeClient {
         // goes awry.
         try {
             // Guards the creation of a staging repo on the release property.
-            // Ensures that we don't create spurious repositories.
+            // Ensures that we don't create spurious repositories if we're not
+            // running a release.
             if (project.properties.containsKey("release")) {
                 // Avoids creating the staging repository more than once.
                 // Implements double checked locking to initialize STAGING_REPO_URL.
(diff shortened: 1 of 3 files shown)


devchas (approved):

buildSrc/src/main/groovy/com.google.api-ads.nexus-publish.gradle
```diff
@@ -0,0 +1,274 @@
...
+ * Logging of raw HTTP is handled by the the HTTP client (google-java-http-client) which uses JUL.
+ *
+ * This implementation is based on:
+ * https://support.sonatype.com/hc/en-us/articles/213465868-Uploading-to-a-Staging-Repository-via-REST-API
+ *
+ * Note that there are two versions of Nexus: v2 and v3. The OSS instance (which hosts our project)
+ * uses v2.
+ *
+ * The Nexus docs are here: https://help.sonatype.com/repomanager2
+ * The REST docs are here: https://repository.sonatype
```
jradcliff, line 133: nit: Fix line break

buildSrc/src/main/groovy/com.google.api-ads.nexus-publish.gradle
```diff
@@ -0,0 +1,274 @@
...
+            logger.warn("Found multiple staging profiles, using the first one")
+        }
+        def stagingStartUrl = new GenericUrl(stagingProfileUrls[0] + "/start")
+        def content = new PromoteRequestXmlContent(
+                NAMESPACE_DICTIONARY, generateStagingRepoDescription())
+        def response = REQUEST_FACTORY
+                .buildPostRequest(stagingStartUrl, content)
+                .setHeaders(authHeaders)
+                .execute()
+        // Checks the response and returns the repository URL, falling back to auto-staging if
```
jradcliff, line 217: nit: comment on fallback is referring to the caller behavior. This should say it throws an exception.

jradcliff (commented):
Two small nits on comments

PR #551, opened 2022-01-12
Override the log4j dependency used by the 'shadow' plugin

jradcliff:


PierrickVoulet (approved):

PierrickVoulet:
@jradcliff Any way to upgrade the version of the plugin itself instead?

The current version of this PR works just fine but this may lead to some additional maintenance costs later on.

jradcliff:
> @jradcliff Any way to upgrade the version of the plugin itself instead?
> 
> The current version of this PR works just fine but this may lead to some additional maintenance costs later on.

I looked into that, but it's not clear to me if that would require upgrading Gradle itself to 7.x, so I'm making the dependency change while I'm investigating.

jradcliff:
> > @jradcliff Any way to upgrade the version of the plugin itself instead?
> > The current version of this PR works just fine but this may lead to some additional maintenance costs later on.
> 
> I looked into that, but it's not clear to me if that would require upgrading Gradle itself to 7.x, so I'm making the dependency change while I'm investigating.

Confirmed: this would require upgrading to 7.x.

https://github.com/johnrengelman/shadow#latest-test-compatibility

devchas (approved):

PR #593, opened 2022-03-22
Add paged search example

devchas:


google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java
```diff
@@ -0,0 +1,247 @@
...
+ * starting by the first page. More details at
+ * https://developers.google.com/google-ads/api/docs/reporting/paging.
+ *
+ * <p>This example searches campaigns illustrating how GoogleAdsService.Search result page tokens
+ * can be cached and reused to retrieve previous pages. This is useful when you need to request
+ * pages that were already requested in the past without starting over from the first page. For
+ * example, it can be used to implement an interactive application that displays a page of results
+ * at a time without caching all the results first.
+ *
+ * <p>To add campaigns, {@link com.google.ads.googleads.examples.basicoperations.AddCampaigns}.
```
jradcliff, line 45: missing "run" before the link

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java
```diff
@@ -0,0 +1,247 @@
...
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * GoogleAdsService.Search results are paginated, but they can only be retrieved in sequence
+ * starting by the first page. More details at
+ * https://developers.google.com/google-ads/api/docs/reporting/paging.
+ *
+ * <p>This example searches campaigns illustrating how GoogleAdsService.Search result page tokens
+ * can be cached and reused to retrieve previous pages. This is useful when you need to request
```
jradcliff, line 40: nit: please use the Java camel case style, with `Search` as `search`.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java
```diff
@@ -0,0 +1,247 @@
...
+      long totalNumberOfResults =
+          Math.min(RESULTS_LIMIT, response.getPage().getResponse().getTotalResultsCount());
+      System.out.printf("Total number of campaigns found: %d.%n", totalNumberOfResults);
+
+      // Determines the total number of pages and prints it.
+      totalNumberOfPages = (int) Math.ceil((double) totalNumberOfResults / (double) PAGE_SIZE);
+      System.out.printf("Total number of pages: %d.%n", totalNumberOfPages);
+      if (totalNumberOfPages == 0) {
+        System.out.println("Could not find any campaigns.");
+        System.exit(1);
```
jradcliff, line 147: Prefer `return` over `System.exit`. We should only do the latter in `main` so if someone copies and pastes this code they won't inadvertently shut down their JVM on an empty response.

If detecting an empty response is important (e.g., for FT), then you could just throw an exception here such as `IllegalStateException`.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java
```diff
@@ -0,0 +1,247 @@
...
+      }
+    }
+
+    // Demonstrates how the logic works when iterating pages forward. We select a page that is
+    // in the middle of the result set so that only a subset of the page tokens will be cached.
+    int middlePageNumber = (int) Math.ceil((double) totalNumberOfPages / 2.0);
```
jradcliff, line 153: You don't need the `(double)` since you're dividing by `2.0`.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java
```diff
@@ -0,0 +1,247 @@
...
+
+    // Demonstrates how the logic works when iterating pages backward with some page tokens that
+    // are not already cached.
+    System.out.println("--- 2. Print results from the last page to the first:");
+    for (int pageNumber = totalNumberOfPages; pageNumber > 0; pageNumber--) {
+      System.out.printf("-- Printing results for page %d:%n", pageNumber);
+      fetchAndPrintPageResults(googleAdsClient, customerId, query, pageNumber, pageTokens);
+    }
+  }
+
```
jradcliff, line 165: In terms of the output a user will see on the console, I think just "Results of the xxx:" is better. "Print results..." reads like code comments.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java
```diff
@@ -0,0 +1,247 @@
...
+    int totalNumberOfPages;
+    try (GoogleAdsServiceClient googleAdsServiceClient =
+        googleAdsClient.getLatestVersion().createGoogleAdsServiceClient()) {
+      // Issues a paginated search request.
+      SearchPagedResponse response = googleAdsServiceClient.search(request);
+      cacheNextPageToken(pageTokens, response.getPage(), 1);
+
+      // Determines the total number of results and prints it.
+      // The total results count does not take into consideration the LIMIT clause of the query,
```
jradcliff, line 136: Similar to the other PR, could you replace the "we" phrases to match the style we use elsewhere for Java?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java
```diff
@@ -0,0 +1,247 @@
...
+   *
+   * @param pageTokens the cache of page tokens to update.
+   * @param page the page that was retrieved.
+   * @param pageNumber the number of the page that was retrieved.
+   */
+  private static void cacheNextPageToken(List<String> pageTokens, SearchPage page, int pageNumber) {
+    if (page.hasNextPage() && pageTokens.size() < pageNumber + 1) {
+      // Updates the cache with the next page token if it is not set yet.
+      pageTokens.add(page.getNextPageToken());
+      // Prints in green color for better console readability.
```
jradcliff, line 243: We're not actually printing in green. Not sure how to do that with Java, TBH. :smile:

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java
```diff
@@ -0,0 +1,247 @@
...
+   * @param googleAdsClient the Google Ads API client.
+   * @param customerId the client customer ID.
+   * @throws GoogleAdsException if an API request failed with one or more service errors.
+   */
+  private static void runExample(GoogleAdsClient googleAdsClient, long customerId) {
+    // The cache of page tokens. It is stored in-memory and in ascendant order of page number.
+    // The first page's token is always an empty string.
+    List<String> pageTokens = new ArrayList<>(Collections.singletonList(""));
+
+    System.out.println("--- 0. Fetch page 1 to get metadata:");
```
jradcliff, line 111: Move this to above line 132

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java
```diff
@@ -0,0 +1,247 @@
...
+
+      // Determines the total number of results and prints it.
+      // The total results count does not take into consideration the LIMIT clause of the query,
+      // so we need to find the minimal value between the limit and the total results count.
+      long totalNumberOfResults =
+          Math.min(RESULTS_LIMIT, response.getPage().getResponse().getTotalResultsCount());
+      System.out.printf("Total number of campaigns found: %d.%n", totalNumberOfResults);
+
+      // Determines the total number of pages and prints it.
+      totalNumberOfPages = (int) Math.ceil((double) totalNumberOfResults / (double) PAGE_SIZE);
```
jradcliff, line 143: Only need to cast one of the two values in the division to `double`.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java
```diff
@@ -0,0 +1,247 @@
...
+   * Runs this example.
+   *
+   * @param googleAdsClient the Google Ads API client.
+   * @param customerId the client customer ID.
+   * @throws GoogleAdsException if an API request failed with one or more service errors.
+   */
+  private static void runExample(GoogleAdsClient googleAdsClient, long customerId) {
+    // The cache of page tokens. It is stored in-memory and in ascendant order of page number.
+    // The first page's token is always an empty string.
+    List<String> pageTokens = new ArrayList<>(Collections.singletonList(""));
```
jradcliff, line 109: Did you consider using a `SortedMap<Integer, String>` for this? I think that would be more intuitive for a cache, and would avoid the mental exercise of translating 0-based list indices to 1-based page numbers. I think that'd make `cacheNextPageToken()` and lines 183-194 simpler.

You'd still be able to iterate over the keys (page numbers) in order with a `SortedMap`, and could also use `headMap` or `tailMap` if you need all keys `< someKey` or all keys `>= someKey`, respectively.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java
```diff
@@ -0,0 +1,247 @@
...
+  // [END navigate_search_result_pages_caching_tokens]
+
+  /**
+   * Updates the cache of page tokens based on a page that was retrieved.
+   *
+   * @param pageTokens the cache of page tokens to update.
+   * @param page the page that was retrieved.
+   * @param pageNumber the number of the page that was retrieved.
+   */
+  private static void cacheNextPageToken(List<String> pageTokens, SearchPage page, int pageNumber) {
```
jradcliff, line 239: Perhaps it's just me, but it feels a bit strange to pass the `pageNumber` of `page` here instead of the page number corresponding to the token that will be cached.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java
devchas: I went back and forth on this using a map, but didn't think to using a SORTED map - thanks!

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java
devchas: Agreed, and I think this makes more sense in the context of a map.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java
```diff
@@ -27,22 +27,21 @@
...
 import java.io.FileNotFoundException;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
+import java.util.SortedMap;
+import java.util.TreeMap;
 
 /**
  * GoogleAdsService.Search results are paginated, but they can only be retrieved in sequence
```
jradcliff, line 34: nit: fix `Search` here as well

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java
```diff
@@ -140,25 +139,25 @@ private static void runExample(GoogleAdsClient googleAdsClient, long customerId)
...
+      totalNumberOfPages = (int) Math.ceil(totalNumberOfResults / (double) PAGE_SIZE);
       System.out.printf("Total number of pages: %d.%n", totalNumberOfPages);
       if (totalNumberOfPages == 0) {
         System.out.println("Could not find any campaigns.");
-        System.exit(1);
+        return;
       }
     }
 
     // Demonstrates how the logic works when iterating pages forward. We select a page that is
```
jradcliff, line 150: One last "we" sentence. How about:

`Demonstrates how the logic works when iterating pages forward. Selects a page...`

jradcliff (commented):
A few minor comments

google-ads-examples/src/main/java/com/google/ads/googleads/examples/misc/NavigateSearchResultPagesCachingTokens.java
devchas: There are actually a bunch of these throughout the example. I know I changed the language in the last example, but is this something we feel strongly about? I do think we've used the "we" phrasing in the past, and it may become a little burdensome updating all the comments from golden CLs moving forward if Java is the only library abiding by this.

PierrickVoulet (approved):

jradcliff (approved):

PR #600, opened 2022-03-31
Remove google-ads-stubs-v7 folder

jradcliff:
Support for v7 of the API was removed in release 17.0.0.


PierrickVoulet (approved):

devchas:
Does this commit remove the v7 folder itself or just the files contained within? It's difficult to tell from the UI, but it looks like just the folders within.

jradcliff:
@devchas , it removes both the folder and its files. I think GitHub only shows the files in the diff because it implicitly removes a folder once it's empty.

PR #626, opened 2022-05-19
Add method to extract all errors of a partial failure status at once

markusheiden:
This is useful because getting all errors via their operation index is inefficient. In #getGoogleAdsErrors(long, GoogleAdsFailureT) the error paths for all failures are build just to pick those for the requested operation index. That leads creation of many ErrorPath objects for all not requested operation indexes which. Those are immediately discarded what leads to avoidable GC pressure.

This can already be achieved with the current API, but I think it belongs into the error util. The code (Java 11) I currently use as supplement:
status.getDetailsList().stream()
  .map(errorUtils::getGoogleAdsFailure)
  .map(errorUtils::getGoogleAdsErrors)
  .flatMap(List::stream)
  .collect(toSet());

### DIFF
--- a/google-ads-stubs-lib/src/main/java/com/google/ads/googleads/lib/utils/AbstractErrorUtils.java
+++ b/google-ads-stubs-lib/src/main/java/com/google/ads/googleads/lib/utils/AbstractErrorUtils.java
@@ -31,6 +31,68 @@ public abstract class AbstractErrorUtils<
     FieldPathElementT extends Message> {
 
   /**
+   * Gets a list of all partial failure error messages for a given response.
+   *
+   * <p>For example, given the following Failure:
+   *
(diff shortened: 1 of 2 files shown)


markusheiden:
When implementing this PR I wondered why the errors in `AbstractErrorUtils#getGoogleAdsErrors(long, GoogleAdsFailureT)` are deduplicated via `if (!result.contains(error)) result.add(error)`? Using a `java.util.Set` in this case would IMO be more efficient because the duplicate check performance currently decreases with the length of the `java.util.List` used for `result`. This IMO needs an incompatible API change of `AbstractErrorUtils` though.

SamyDgz (approved):

SamyDgz (approved):

PR #663, opened 2022-08-16
Add GenerateHistoricalMetrics example

devchas:


google-ads-examples/src/main/java/com/google/ads/googleads/examples/planning/GenerateHistoricalMetrics.java
```diff
@@ -0,0 +1,157 @@
+// Copyright 2019 Google LLC
```
jradcliff, line 1: 2022

google-ads-examples/src/main/java/com/google/ads/googleads/examples/planning/GenerateHistoricalMetrics.java
```diff
@@ -0,0 +1,157 @@
...
+    try (KeywordPlanServiceClient keywordPlanServiceClient =
+        googleAdsClient.getLatestVersion().createKeywordPlanServiceClient()) {
+      GenerateHistoricalMetricsResponse response =
+          keywordPlanServiceClient.generateHistoricalMetrics(resourceName);
+
+      for (KeywordPlanKeywordHistoricalMetrics metric : response.getMetricsList()) {
+        // These metrics include those for both the search query and any close variants included in
+        // the response.
+
+        ProtocolStringList variants = metric.getCloseVariantsList();
```
jradcliff, line 111: Can we just declare this as `List<String>`? I think the `ProtocolStringList` type could be distracting, and AFAICT we're not doing anything here that requires methods from that specific subclass of `List<String>`.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/planning/GenerateHistoricalMetrics.java
```diff
@@ -0,0 +1,157 @@
...
+        // The competition level for this search query.
+        System.out.printf(
+            "\tCompetition level: %s.%n", metric.getKeywordMetrics().getCompetition().name());
+
+        // The competition index for the query in the range [0, 100]. This shows how competitive ad
+        // placement is for a keyword. The level of competition from 0-100 is determined by the
+        // number of ad slots filled divided by the total number of ad slots available. If not
+        // enough data is available, None will be returned.
+        System.out.printf(
+            "\tCompetition index: %d.%n", metric.getKeywordMetrics().getCompetitionIndex());
```
jradcliff, line 135: I'd either:

a) conditionally print this based on `hasCompetitionIndex()` or
b) print "none" or "n/a" if not present (this'll require changing from `%d` to `%s`)

since getCompetitionIndex() will just return `0` if the metrics object doesn't have a value for the index.

jradcliff: Same applies to lowTopOfPageBidMicros and highTopOfPageBidMicros if those can be `null`.

jradcliff (approved):

BenRKarl (approved):

PR #697, opened 2023-03-10
Add full example of creating a Responsive Search Ad (with keywords and geo targeting)

jradcliff:
Also includes minor cleanup of reformatting campaignmanagement/CreateExperiment using google-java-format.


jradcliff:
Will revisit this once we've updated the approach for image extensions

PR #729, opened 2023-08-16
Remove support for v12

jradcliff:
This version will sunset on September 27, 2023:

https://developers.google.com/google-ads/api/docs/sunset-dates


devchas (commented):
Do we need to remove the v12 reference in the google-ads-parent settings.gradle file? That was the only reference I could find outside of the v12 directory.

jradcliff:
Good point re: `settings.gradle`. I made that change in the first commit but GitHub is struggling mightily to show it. You can see it directly in the `remove-v12` branch for this PR at https://github.com/googleads/google-ads-java/blob/remove-v12/settings.gradle, howerver.

devchas (approved):

PR #740, opened 2023-10-20
Upgrade to mockito 4.11.0 and hamcrest 2.2

jradcliff:
- Mockito 4.11.0 is the last version of Mockito that's compatible with Java8.
- Hamcrest 2.2 is the latest version and uses a simpler naming and versioning convention. See https://github.com/hamcrest/JavaHamcrest/releases.

Bug: b/303638013
Change-Id: Ia6df059a0701c1bfc2171c7611ddda6da89f8bd9


google-ads/src/test/java/com/google/ads/googleads/lib/logging/LoggingInterceptorTest.java
```diff
@@ -593,9 +593,7 @@ private void resetMocks() {
 
   private void enableAllLevels(Logger logger) {
     when(logger.isDebugEnabled()).thenReturn(true);
-    when(logger.isErrorEnabled()).thenReturn(true);
```
jradcliff, line 596: Removed these mocking calls because they were never used

devchas (approved):

PR #742, opened 2023-10-24
Remove the usage of eCPC bidding because of its deprecation

yangda:
As seen in [this blog announcement](https://ads-developers.googleblog.com/2023/09/google-ads-shopping-campaign-enhanced.html), any Shopping campaigns using Enhanced cost-per-click (eCPC) will behave as if they are using Manual cost-per-click (CPC) bidding.

Because of the change, I'm updating the examples related to the announcement to reflect the new recommended practice.

### DIFF
--- a/google-ads-examples/src/main/java/com/google/ads/googleads/examples/shoppingads/AddShoppingProductAd.java
+++ b/google-ads-examples/src/main/java/com/google/ads/googleads/examples/shoppingads/AddShoppingProductAd.java
@@ -237,11 +237,11 @@ public class AddShoppingProductAd {
             // the ads from immediately serving. Set to ENABLED once you've added
             // targeting and the ads are ready to serve.
             .setStatus(CampaignStatus.PAUSED)
-            // Sets the bidding strategy to Manual CPC (with eCPC enabled)
+            // Sets the bidding strategy to Manual CPC (with eCPC disabled)
             // Recommendation: Use one of the automated bidding strategies for Shopping campaigns
             // to help you optimize your advertising spend. More information can be found here:
             // https://support.google.com/google-ads/answer/6309029.
-            .setManualCpc(ManualCpc.newBuilder().setEnhancedCpcEnabled(true).build())
+            .setManualCpc(ManualCpc.newBuilder().setEnhancedCpcEnabled(false).build())
             // Sets the budget.
             .setCampaignBudget(budgetResourceName)
             .build();
--- a/google-ads-examples/src/main/java/com/google/ads/googleads/examples/shoppingads/AddShoppingProductAd.java
+++ b/google-ads-examples/src/main/java/com/google/ads/googleads/examples/shoppingads/AddShoppingProductAd.java
@@ -237,7 +237,10 @@ public class AddShoppingProductAd {
             // the ads from immediately serving. Set to ENABLED once you've added
             // targeting and the ads are ready to serve.
             .setStatus(CampaignStatus.PAUSED)
-            // Sets the bidding strategy to Manual CPC (with eCPC disabled)
+            // Sets the bidding strategy to Manual CPC (with eCPC disabled). eCPC for standard
+            // Shopping campaigns is deprecated. If eCPC is set to true, Google Ads ignores the
+            // the setting and behaves as if the setting was false. See this blog post for more
+            // information: https://ads-developers.googleblog.com/2023/09/google-ads-shopping-campaign-enhanced.html
             // Recommendation: Use one of the automated bidding strategies for Shopping campaigns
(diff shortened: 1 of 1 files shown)


google-ads-examples/src/main/java/com/google/ads/googleads/examples/shoppingads/AddShoppingProductAd.java
```diff
@@ -237,11 +237,11 @@ private String addStandardShoppingCampaign(
             // the ads from immediately serving. Set to ENABLED once you've added
             // targeting and the ads are ready to serve.
             .setStatus(CampaignStatus.PAUSED)
-            // Sets the bidding strategy to Manual CPC (with eCPC enabled)
+            // Sets the bidding strategy to Manual CPC (with eCPC disabled)
```
jradcliff, line 240: Could you add a comment here explaining that if they set it to `true`, Google Ads will ignore the setting and the campaign will behave as if the setting was `false`, and link to the HC or blog post explaining why?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/shoppingads/AddShoppingProductAd.java
```diff
@@ -237,11 +237,11 @@ private String addStandardShoppingCampaign(
...
             // Recommendation: Use one of the automated bidding strategies for Shopping campaigns
             // to help you optimize your advertising spend. More information can be found here:
             // https://support.google.com/google-ads/answer/6309029.
-            .setManualCpc(ManualCpc.newBuilder().setEnhancedCpcEnabled(true).build())
+            .setManualCpc(ManualCpc.newBuilder().setEnhancedCpcEnabled(false).build())
```
jradcliff, line 244: Should we be setting this at all? What will happen if the user just does `.setManualCpc(ManualCpc.newBuilder())`?

yangda: I thought about this while working on it, and observed that if we do not set enhanced_cpc_enabled explicitly, by default the value is `true` by checking the result for the created campaign's settings in the Google Ads UI. With that, I feel it might be better to encourage setting `false` explicitly in this example to avoid confusions (not setting eCPC enabled in code -> created campaign with eCPC enabled -> the actual bidding behavior is with eCPC disabled).

P.S. More info on my test: I did it with the python client lib by changing the according line to `client.copy_from(campaign.manual_cpc, client.get_type("ManualCpc"))`. I think that should equal to the Java `.setManualCpc(ManualCpc.newBuilder().build())`, by referring to related code in [this example](https://developers.google.com/google-ads/api/samples/add-complete-campaigns-using-batch-job#python).

jradcliff: Thanks for checking. Given it defaults to `true`, I agree with the approach of explicitly setting it to `false`.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/shoppingads/AddShoppingProductAd.java
yangda: Sure! Added suggested comments in a new commit.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/shoppingads/AddShoppingProductAd.java
```diff
@@ -237,11 +237,14 @@ private String addStandardShoppingCampaign(
             // the ads from immediately serving. Set to ENABLED once you've added
             // targeting and the ads are ready to serve.
             .setStatus(CampaignStatus.PAUSED)
-            // Sets the bidding strategy to Manual CPC (with eCPC enabled)
+            // Sets the bidding strategy to Manual CPC (with eCPC disabled). eCPC for standard
+            // Shopping campaigns is deprecated. If eCPC is set to true, Google Ads ignores the
+            // the setting and behaves as if the setting was false. See this blog post for more
+            // information: https://ads-developers.googleblog.com/2023/09/google-ads-shopping-campaign-enhanced.html
```
jradcliff, line 243: Since URLs (including this one) almost always make a line exceed the 100 character standard, we usually put them in a separate comment line.

yangda: Thanks for the info! Done moving it to a separate line.

jradcliff (approved):

jradcliff (approved):

PR #785, opened 2024-03-27
Add the VerifyAdvertiserIdentity example

jradcliff:


Raibaz (approved):

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/VerifyAdvertiserIdentity.java
```diff
@@ -0,0 +1,158 @@
...
+   */
+  private void runExample(GoogleAdsClient googleAdsClient, long customerId) {
+    try (IdentityVerificationServiceClient identityVerificationServiceClient =
+        googleAdsClient.getLatestVersion().createIdentityVerificationServiceClient()) {
+      // Retrieves the current Advertiser Identity Verification status.
+      IdentityVerification identityVerification =
+          getIdentityVerification(customerId, identityVerificationServiceClient);
+
+      if (identityVerification != null) {
+        if (identityVerification.hasVerificationProgress()
```
devchas, line 95: I know you included the `hasVerificationProgress` check as a safeguard against the `getVerificationProgress` call, but is there a possible scenario where `indentityVerification != null && hasVerificationProgress == false`? In that case, you'd get routed to display the URL, but wouldn't you want to `startIdentityVerification` instead? Without knowing more about the feature itself, I'm not sure if that scenario is possible, but I wanted to flag it for you.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/VerifyAdvertiserIdentity.java
```diff
@@ -0,0 +1,158 @@
...
+        } else {
+          // If there is an identity verification session in progress, there is no need to start
+          // another one by calling startIdentityVerification.
+          System.out.println("There is an advertiser identify verification session in progress.");
+          System.out.printf(
+              "The URL for the verification process is '%s' and it will expire at '%s'%n",
+              identityVerification.getVerificationProgress().getActionUrl(),
+              identityVerification.getVerificationProgress().getInvitationLinkExpirationTime());
+        }
+      } else {
```
devchas, line 112: Optional style comment: I think you can restructure this to (assuming nothing changes based on my other comment)
```
if (identityVerification == null)
...
else if (...actionUrl is empty)
...
else
...
```

This creates three distinct states without a nested if statement, but your call.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/VerifyAdvertiserIdentity.java
jradcliff: @Raibaz , could you comment on what we should do in the scenario where there is an `IdentityVerification` but it either:

a) Doesn't have a `VerificationProgress`

or

b) Has a `VerificationProgress`, but the progress doesn't have an `actionUrl`?

jradcliff: Thanks for pointing out all of this. The latest commit uses a `switch` statement as discussed over email.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/VerifyAdvertiserIdentity.java
jradcliff: Addressed with new approach of `switch` on program status

devchas (approved):

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/VerifyAdvertiserIdentity.java
```diff
@@ -0,0 +1,173 @@
...
+            .getIdentityVerificationRequirement()
+            .getVerificationCompletionDeadlineTime();
+    IdentityVerificationProgress progress = identityVerification.getVerificationProgress();
+    System.out.printf(
+        "Account %d has a verification completion deadline of '%s' and status '%s' for advertiser"
+            + " identity verification.%n",
+        customerId, deadline, progress.getProgramStatus());
+    return identityVerification;
+  }
+
```
Raibaz, line 161: Nit: needless empty line

Raibaz (approved):

google-ads-examples/src/main/java/com/google/ads/googleads/examples/accountmanagement/VerifyAdvertiserIdentity.java
jradcliff: For some reason, the Google Java Formatter insists on adding this empty line whenever I try to take it out. I'd prefer to leave it in so we don't have to manually fix it any time we touch this file going forward.

Raibaz: 🤷🏻

PR #788, opened 2024-03-29
Modify conversion adjustment example to use order_id

jradcliff:
Although the API still supports uploading using GCLID and conversion date/time, using order ID is strongly preferred.

https://support.google.com/google-ads/answer/7686280

### DIFF
--- a/google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadConversionAdjustment.java
+++ b/google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadConversionAdjustment.java
@@ -25,12 +25,14 @@ import com.google.ads.googleads.v16.errors.GoogleAdsFailure;
 import com.google.ads.googleads.v16.services.ConversionAdjustment;
 import com.google.ads.googleads.v16.services.ConversionAdjustmentResult;
 import com.google.ads.googleads.v16.services.ConversionAdjustmentUploadServiceClient;
-import com.google.ads.googleads.v16.services.GclidDateTimePair;
 import com.google.ads.googleads.v16.services.RestatementValue;
 import com.google.ads.googleads.v16.services.UploadConversionAdjustmentsRequest;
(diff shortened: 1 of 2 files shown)


devchas (approved):

google-ads-examples/src/main/java/com/google/ads/googleads/examples/remarketing/UploadConversionAdjustment.java
```diff
@@ -157,26 +154,29 @@ private void runExample(
...
             .setConversionAction(ResourceNames.conversionAction(customerId, conversionActionId))
             .setAdjustmentType(conversionAdjustmentType)
-            .setGclidDateTimePair(
-                GclidDateTimePair.newBuilder()
-                    .setGclid(gclid)
-                    .setConversionDateTime(conversionDateTime)
-                    .build())
+            // Sets the orderId to identify the conversion to adjust.
+            .setOrderId(orderId)
+            // As an alternative to setting orderId, you can provide a GclidDateTimePair, but
```
BenRKarl, line 173: Docs state that `gclid_date_time_pair is optional but may be set in addition to the order_id.` so this comment should be rephrased to reflect that. I'm not sure what the best-case would be, but I imagine the recommendation would be to provide both gclid date time _and_ an order ID. Josh perhaps you know better than I do on that.

BenRKarl: Just realized that we want to steer users towards order ID and away from gclid date/time pairs. So I think the comment should note that both can be set, but that we recommend always setting order ID.

jradcliff: I confirmed separately that we should not recommend passing both `order_id` and `gclid_date_time_pair`, so I'll leave this section as-is.

BenRKarl (approved):

PR #816, opened 2024-12-16
Remove support for v16

sarahcaseybot:
This version will sunset on February 5, 2025

The full deprecation and sunset schedule can be found [here](https://developers.google.com/google-ads/api/docs/sunset-dates).


google-ads-stubs-v16/build.gradle
jradcliff, line 1: Please also remove `google-ads-stubs-v16` from https://github.com/googleads/google-ads-java/blob/main/settings.gradle

jradcliff (changes requested):

google-ads-stubs-v16/build.gradle
sarahcaseybot: I already did this in the [initial commit](https://github.com/googleads/google-ads-java/blob/9152af6a038811dbe7579ed8713beb1f2451fab5/settings.gradle). I believe it's getting cut off from the "files changed" tab since there are so many changed files in this PR.

jradcliff: Yep, you're right, sorry about that! I diffed on the command line and LGTM.

jradcliff (approved):

PR #830, opened 2025-05-12
Remove v17

sarahcaseybot:
Sunset date for v17 is June 4, 2025


jradcliff (changes requested):
Please also remove the `include` for v17 from `settings.gradle`

jradcliff (approved):

PR #865, opened 2025-11-25
Add ADC environment variable

sarahcaseybot:


jradcliff:
I see this adds the `ConfigPropertyKey`, but are more changes needed to actually *use* the key in client construction?

PR #887, opened 2026-06-03
Add experiments examples

sarahcaseybot:


google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/CreateSearchCustomExperiment.java
```diff
@@ -47,12 +47,19 @@
...
 /**
- * This example creates a new experiment, experiment arms, and demonstrates how to modify the draft
- * campaign as well as begin the experiment.
+ * This example creates a standard, system-managed campaign experiment.
+ *
+ * <p>It demonstrates how to create an experiment, configure its control and treatment arms (where
+ * the treatment arm automatically generates a draft campaign), modify the system-generated draft
+ * campaign, and schedule the experiment.
+ *
+ * <p>Note: This standard draft-based workflow does not apply to all experiment types (e.g.,
```
jradcliff: nit: avoid e.g., per https://developers.google.com/style/word-list#eg

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/GetExperimentPerformance.java
```diff
@@ -0,0 +1,266 @@
...
+      SearchGoogleAdsRequest request =
+          SearchGoogleAdsRequest.newBuilder()
+              .setCustomerId(Long.toString(customerId))
+              .setQuery(query)
+              .build();
+
+      GoogleAdsServiceClient.SearchPagedResponse response = googleAdsServiceClient.search(request);
+
+      boolean experimentFound = false;
+      for (GoogleAdsRow row : response.iterateAll()) {
```
jradcliff: Can you get multiple rows back, given you're filtering by `experiment_id`? If not, it's a little confusing to have a `for` loop here. You could use something like https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/Iterables.html#getOnlyElement(java.lang.Iterable) instead.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/GetExperimentPerformance.java
```diff
@@ -0,0 +1,266 @@
...
+      }
+
+      if (!experimentFound) {
+        System.out.printf("No experiment found for experiment ID: %d%n", experimentId);
+      }
+    }
+  }
+
+  // [START get_experiment_performance_1]
+  private void evaluateExperiment(GoogleAdsClient googleAdsClient, long customerId, GoogleAdsRow row) {
```
jradcliff: Please add javadocs for each method

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/GetExperimentPerformance.java
```diff
@@ -0,0 +1,266 @@
...
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.UUID;
+
+/**
+ * This example illustrates how to retrieve performance metrics for an experiment.
+ *
+ * <p>It shows how to query statistical significance metrics for the experiment,
+ * and how to execute actions such as promoting, ending, or graduating an experiment.
```
jradcliff: The name and description of this example don't make it obvious that running this will potentially modify the state of the experiment.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/EvaluateAndUpdateExperiment.java
```diff
@@ -0,0 +1,266 @@
...
+  private void endExperiment(GoogleAdsClient googleAdsClient, long customerId, String experimentResourceName) {
+    try (ExperimentServiceClient experimentServiceClient =
+        googleAdsClient.getLatestVersion().createExperimentServiceClient()) {
+      experimentServiceClient.endExperiment(experimentResourceName);
+      System.out.printf("Successfully ended experiment: %s%n", experimentResourceName);
+    }
+  }
+
+  private void graduateExperiment(GoogleAdsClient googleAdsClient, long customerId, String experimentResourceName) {
+    String budgetResourceName;
```
jradcliff, line 260: Some line-level comments would be helpful since this method performs multiple steps

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/CreateSearchCustomExperiment.java
```diff
@@ -47,12 +47,19 @@
...
+ * intra-campaign or asset optimization experiments) that do not use system-generated treatment campaign
+ * copies.
  */
-public class CreateExperiment {
+public class CreateSearchCustomExperiment {
```
jradcliff, line 60: This has the term "Custom" but lines 50 and 56 suggest this is a standard experiment. Are we using "Custom" here because it's customizing a standard experiment, maybe?

laurachevalier4: It's because the experiment type is SEARCH_CUSTOM.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/GetExperimentPerformance.java
laurachevalier4: That's a good point. What about something like EvaluateAndUpdateExperiment or something like that? 

For the description, do you mean that the first sentence should better capture what the example does, or you think the whole description is incomplete?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/CreateSearchCustomExperiment.java
laurachevalier4: I can fix this in the golden.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/EvaluateAndUpdateExperiment.java
laurachevalier4: I can fix this in the golden. @sarahcaseybot I'll follow up with you on that.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/GetExperimentPerformance.java
jradcliff: > That's a good point. What about something like EvaluateAndUpdateExperiment or something like that?

`EvaluateAndUpdateExperiment` SGTM. I just want it to be obvious to a user that running this is going to mutate things in their account.

> For the description, do you mean that the first sentence should better capture what the example does

Exactly. It should mention the mutate piece. Javadoc uses the first sentence of the description in method summaries and IDEs use those summaries in tooltips, so it's important that the first sentence mentions the sample will modify entities.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/EvaluateAndUpdateExperiment.java
laurachevalier4: Merged the changes in Python: https://github.com/googleads/google-ads-python/pull/1078

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/GetExperimentPerformance.java
sarahcaseybot: Done.

jradcliff (approved):

laurachevalier4 (approved):

PR #889, opened 2026-06-09
Add asset optimization experiment example

sarahcaseybot:


google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/CreateAssetOptimizationExperiment.java
```diff
@@ -0,0 +1,322 @@
...
+   * Runs the example.
+   *
+   * @param googleAdsClient the googleAdsClient.
+   * @param customerId the customer ID.
+   * @param assetGroupId the base asset group ID to run the experiment on.
+   */
+  private void runExample(GoogleAdsClient googleAdsClient, long customerId, long assetGroupId)
+      throws IOException {
+
+    String campaignResourceName = null;
```
jradcliff: nit: You can move this into the `try` block since it's only used in that scope

jradcliff (approved):

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/CreateAssetOptimizationExperiment.java
```diff
@@ -0,0 +1,322 @@
...
+      // Send all operations in a single Mutate request.
+      // The operations must be in this specific order.
+      MutateGoogleAdsRequest mutateRequest =
+          MutateGoogleAdsRequest.newBuilder()
+              .setCustomerId(Long.toString(customerId))
+              .addMutateOperations(textAssetOperation)
+              .addMutateOperations(imageAssetOperation)
+              .addMutateOperations(experimentOperation)
+              .addMutateOperations(assetGroupAssetOperation1)
+              .addMutateOperations(assetGroupAssetOperation2)
```
laurachevalier4, line 290: The asset group asset operations should go last, after the control / treatment arm operations.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/CreateAssetOptimizationExperiment.java
```diff
@@ -0,0 +1,322 @@
...
+                  ExperimentArmOperation.newBuilder().setCreate(treatmentArm).build())
+              .build();
+
+      // 4. Create AssetGroupAssets linking the assets to the asset group.
+      AssetGroupAsset assetGroupAsset1 =
+          AssetGroupAsset.newBuilder()
+              .setAssetGroup(assetGroupResourceName)
+              .setAsset(asset1TempId)
+              .setFieldType(AssetFieldType.HEADLINE)
+              .setStatus(AssetLinkStatus.PAUSED)
```
laurachevalier4, line 258: This (setting the status to PAUSED) isn't in the golden. Is it necessary in Java? Have you tested whether it works?

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/CreateAssetOptimizationExperiment.java
sarahcaseybot: When I ran the example with the operations in that order I get the failure NOT_ENOUGH_MARKETING_IMAGE_ASSET. According to Jetski "the request fails due to how the Google Ads backend synchronizesmutations during the transaction.
When the Treatment Arm is created, we provide the temporary asset -3 as a MARKETING_IMAGE override, which creates it with an ENABLED status in the Trial Arm. However, if we follow that up by adding the exact same -3 asset to the base Asset Group with a PAUSED status, the backend immediately synchronizes that change to the Trial Arm. This causes the Trial Arm's marketing image to become PAUSED, leaving it with zero enabled marketing images and triggering the NOT_ENOUGH_MARKETING_IMAGE_ASSET validation error."

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/CreateAssetOptimizationExperiment.java
sarahcaseybot: This is also neccesarry to prevent the NOT_ENOUGH_MARKETING_IMAGE_ASSET validation error I mentioned above

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/CreateAssetOptimizationExperiment.java
sarahcaseybot: done.

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/CreateAssetOptimizationExperiment.java
laurachevalier4: You're right! I thought I had tested it but I'm getting the same error you were getting. Reordering the operations to put the experiment arm operations last worked for me. I didn't need to make any additional changes. https://github.com/googleads/google-ads-python/pull/1079

google-ads-examples/src/main/java/com/google/ads/googleads/examples/experiments/CreateAssetOptimizationExperiment.java
laurachevalier4: Interesting, I didn't need to include this for it to work in Python. If it works without this, let's remove it for consistency. Otherwise I think it's fine to keep.

laurachevalier4 (approved):

